PR #4, opened 2018-10-10
Feature/real time graphing

coledie:
Readme is readme.txt


Todo.txt
```diff
@@ -0,0 +1,3 @@
+- Sometimes get error after 10min+ -> ValueError: shape mismatch: objects cannot be broadcast to a single shape
```
ClayMav: Having the Todo.txt still makes me feel like there are still things to be done, is this file really necessary?

metric.py
```diff
@@ -0,0 +1,237 @@
...
+
+
+class Metric:
+    """
+    Piece of data to be tracked.
+
+    Parameters
+    ----------
+    line: axis
+        The animation line to plot data on
```
ClayMav: Does `line` have a datatype? You list `string` for the strings below

metric.py
```diff
@@ -0,0 +1,237 @@
...
+    func: string, optional A
+        The function to generate values(takes variables x, y, z corresponding to data streams)
```
ClayMav: This parameter says it takes a `string`, but the description makes me think it takes a function, which is it?

metric.py
```diff
@@ -0,0 +1,237 @@
...
+    x_stream: string, optional A
+        Data stream for x variable in function
```
ClayMav: Data stream is what?

metric.py
```diff
@@ -0,0 +1,237 @@
...
+    y_stream: string, optional
+        Data stream for y variable in function
```
ClayMav: Same as above

metric.py
```diff
@@ -0,0 +1,237 @@
...
+    z_stream: string, optional
+        Data stream for z variable in function
```
ClayMav: Same as above

metric.py
```diff
@@ -0,0 +1,237 @@
...
+        self._line = line
+
+        self._label = line.get_label() if not label else label
+
+        #
+        # Process xml tag
+        if xml_tag is not None:
+            func = xml_tag.get("func")
+
+            x_stream = xml_tag.get('x_stream')
```
ClayMav: If the user inputs an `x_stream`, won't this overwrite it?

metric.py
```diff
@@ -0,0 +1,237 @@
...
+            y_stream = xml_tag.get('y_stream')
+            z_stream = xml_tag.get('z_stream')
+
+        #
+        # Set func
+        assert func is not None, "No function given to metric!"
```
ClayMav: If func is required, don't default it to None

metric.py
```diff
@@ -0,0 +1,237 @@
...
+        self._x_stream = x_stream
+        self._y_stream = y_stream
+        self._z_stream = z_stream
+
+        #
+        # Create data storage
+        self._data = []
+
+    @property
+    def get_line(self):
```
ClayMav: If it's a property, you should be referencing it not as a getter, but as a property. i.e. `object.line` instead of `object.get_line`

metric.py
```diff
@@ -0,0 +1,237 @@
...
+        Returns
+        -------
+        line
+            The animation line where this metrics data will be plotted
+        """
+
+        return self._line
+
+    @property
+    def get_label(self):
```
ClayMav: Same as above

metric.py
```diff
@@ -0,0 +1,237 @@
...
+
+        Parameters
+        ----------
+        label: string
+            New line label
+        """
+
+        self._label = label
+
+    @property
```
ClayMav: Same as previous property comment

metric.py
```diff
@@ -0,0 +1,237 @@
...
+        Returns
+        -------
+        lambda
+            The lambda functions that this metrics values will be generated with
+        """
+
+        return self._func
+
+    @property
+    def get_x_stream(self):
```
ClayMav: Same as above

metric.py
```diff
@@ -0,0 +1,237 @@
...
+        Returns
+        -------
+        string
+            Data stream to be used for x value in metric.
+        """
+
+        return self._x_stream
+
+    @property
+    def get_y_stream(self):
```
ClayMav: Same as above

metric.py
```diff
@@ -0,0 +1,237 @@
...
+        Returns
+        -------
+        string
+            Data stream to be used for y value in metric.
+        """
+
+        return self._y_stream
+
+    @property
+    def get_z_stream(self):
```
ClayMav: Same as above

metric.py
```diff
@@ -0,0 +1,237 @@
...
+        Parameters
+        ----------
+        stream: string
+            New z_stream value
+        """
+
+        self._z_stream = stream
+
+    @property
+    def get_data(self):
```
ClayMav: Same as above

metric.py
```diff
@@ -0,0 +1,237 @@
...
+
+        Returns
+        -------
+        list
+            All previously generated data
+        """
+
+        return self._data
+
+    @get_data.setter
```
ClayMav: Is this really a setter? This seems to add data, not set a variable.

readme.txt
```diff
@@ -0,0 +1,43 @@
+The grapher manages 1+ individual metrics which are just things that take in specific data streams and process them to
```
ClayMav: General comment, this is a very unhelpful README. There is little structure here. I will point you to a project made by a student here at S&T, he defines a really simple and effective way to structure a README. You need not follow it exactly, but you need to clean this up. With no structure, I have no idea what I am supposed to do with any of the information you are giving me. [https://makeread.me/](https://makeread.me/)

ClayMav: Also, file should be called README.md

real_time_graphing.py
```diff
@@ -0,0 +1,297 @@
...
+from multiprocessing import Queue
+import queue
+
+from time import sleep, time
+from timer import timeit
+
+from metric import Metric
+
+
+def get_demo_data():
```
ClayMav: This function should be in a separate file for testing the system. Make sure not to leave testing functions or code/comments in the production code

real_time_graphing.py
```diff
@@ -0,0 +1,297 @@
...
+        plt.show()
+
+        #
+        # Cleanup
+        self.thread_stop.set()
+
+        for thread in threads.values():
+            thread.join()
+
+    @timeit
```
ClayMav: Is this required for the program?

real_time_graphing.py
```diff
@@ -0,0 +1,297 @@
...
+    def read_config(self):
```
ClayMav: I would suggest using the word parse

real_time_graphing.py
```diff
@@ -0,0 +1,297 @@
...
+                ax.set_ylabel(graph.get('ylabel'))
+
+                for metric in graph.findall('metric'):
+                    color = metric.get('color') if metric.get('color') else next(color_gen)
+
+                    m_line, = ax.plot([], [], color=color, label=metric.get('label'))
+
+                    self.tracked_data.append(Metric(line=m_line, xml_tag=metric))
+
+                if (graph.get('legend') if graph.get('legend') else 'yes') == 'yes':
```
ClayMav: wtf

There must be a better way to do this.

real_time_graphing.py
```diff
@@ -0,0 +1,297 @@
...
+        Reads data from network and puts it in a queue to be processed.
+
+        Parameters
+        ----------
+        thread_queue: Thread queue
+            Thread queue
+        """
+
+        while not self.thread_stop.is_set():
+            data = get_demo_data()
```
ClayMav: It looks to me like the real time graphing will only work on fake data right now. definitely make this general

real_time_graphing.py
```diff
@@ -0,0 +1,297 @@
...
+                    z = data[metric.get_z_stream] if metric.get_z_stream else None
+
+                    x_val = func(x, y, z) if z else (func(x, y) if y else func(x))
+
+                    metric.push_data = x_val
+                self.data_count += 1
+            except queue.Empty:
+                pass
+
+    @timeit
```
ClayMav: Necessary?

real_time_graphing.py
```diff
@@ -0,0 +1,297 @@
...
+
+                    x = data[metric.get_x_stream] if metric.get_x_stream else None
+                    y = data[metric.get_y_stream] if metric.get_y_stream else None
+                    z = data[metric.get_z_stream] if metric.get_z_stream else None
+
+                    x_val = func(x, y, z) if z else (func(x, y) if y else func(x))
+
+                    metric.push_data = x_val
+                self.data_count += 1
+            except queue.Empty:
```
ClayMav: Instead, make this part of the while loop condition

timer.py
```diff
@@ -0,0 +1,32 @@
+# For timing individual functions
```
ClayMav: Is this important to the running of the code? If so, then you can keep it that way, but I would ask that you make sure it is general and can be configured by the user.

ClayMav (changes requested):
Biggest issue is that all of the files are in the root directory. It is going to fudge the git history, but make a folder, call it `tools` or something clever, and make another folder `graphing` inside it. put everything in there. The readme needs to be renamed and massively revamped, and definitely make it so that this can be used generally (i.e. with any drone, and not just on fake data)

Here's to many more pull requests.
You guys did good, just gotta clean it up a bit.

ClayMav (approved):
I will give this the go ahead, but I would like to see the `real_time_graphing.py` get support for real input. A generalized route for info to come into the graph. Anyways, good work, and thanks for the quick fixes after the initial review. Very good work.

ClayMav:
If you have anything else you want to add, do it now, else go ahead and merge it when ready.

PR #5, opened 2018-11-07
Base tools

coledie:
Base usable functionality for logger, csv grapher and the config maker.


ClayMav:
I don't really know how to run any of this, also how does the README.md get deleted like every time you commit? The `.gitignore` seems to have been gutted, it would be better to add on to the file. In addition to this, the azure configuration is gone, not sure if thats intentional. Can you make a README is the tools directory that describes what each of the subdirectories do, and also each subdirectory should have a readme as well so that you can have instructions on how to use the files and such. After you fix that stuff I will give it a technical review.

ClayMav (changes requested):
Address my comment and I will give another look

coledie (commented):
Made changes.

coledie:
https://stackoverflow.com/questions/15373493/python-tkinter-menu-bars-dont-display
"Based on some comments you made to one of the answers, you are apparently running this on a Macintosh. The code works fine, but the menu appears in the mac menubar rather than on the window like it does on Windows and Linux. So, there's nothing wrong with your code as far as the menubar is concerned."
All of the tools work on my pc when I try and run them from the IARC directory.

PR #6, opened 2018-11-11
Addidng the drone_console alias as well as the mavlinkbridge service …

25cent9:
…for future on board computers.

Plz **_LGTM_**


config/alias.txt
```diff
@@ -0,0 +1,2 @@
+alias drone_console="sudo service mavlinkbridge stop && sudo mavproxy.py --master=/dev/serial/by-id/usb-3D_Robotics_PX4_FMU_v2.x_0-if00"
```
ClayMav: Can you instead make this file into a `dronerc` file

ClayMav (changes requested):
Can you make a readme for `config/` with explanations for the purpose and use of each thing in here, as well as directions to use each. For the `dronerc` file the directions should be to use `source dronerc` to apply the alias

ClayMav:
Sorry tiny commit took 400 years to get reviewed, I forgot it was here.

config/alias.txt
25cent9: Done

25cent9:
Made changes accordingly.

ClayMav (approved):
👍

PR #10, opened 2019-01-05
Implementation for drone, drone controller, tasks, instructions, and safety

Thomas-McKanna:
This pull request contains a solid foundation for autonomous flight by providing the following features:
* Drone object which acts as an interface to the drone's flight data and abstracts away the following actions: takeoff, land, move, hover, connect to ardupilot
* Abstract Task object which formalizes how logic is written and executed for various drone actions; includes concrete implementations for takeoff, land, move, and hover
* DroneController object which manages Task objects and continually checks for unsafe conditions or keyboard interrupt (such conditions will trigger an emergency landing)
* Abstract Instruction object which will be used to decode messages sent over the network from the swarm controller and turn those instructions into Tasks
* Various utility files for constants, custom exceptions, wrapper functions for DroneKit, and a shared lock for threading.

**Please see the readme of the pull request for additional details**.


Tasks/task_base.py
```diff
@@ -0,0 +1,68 @@
...
+class TaskBase():
+    """
+    Responsible for implementing the core logic of the various actions
+    that a drone can take (ex. Movement, Follow, Heal, Decode). Must, at
+    a minimum, implement do(), is_done(), and exit_task().
+    """
+    __metaclass__ = abc.ABCMeta
+
+    @abc.abstractmethod
+    def do(self):
```
25cent9: Since `do` is typically a reserved word, can we look into changing `do()` to be `perform()` or `execute()`? That way the naming convention still makes sense.

Tasks/task_base.py
```diff
@@ -0,0 +1,68 @@
...
+
+        Returns:
+        ----------
+        True if the task is done with its goal, and false otherwise
+        (But do not use to query if the task is done, use is_done() instead)
+        """
+        pass
+
+    @abc.abstractmethod
+    def is_done(self):
```
25cent9: This isn't super critical but since this function just returns the value of `self.done` I would suggest looking into making this a property by using `@property`.

25cent9 (dismissed):
**`LGTM`** from me and you can make the changes at your discretion. `Tasks` looks ready to rock and roll.

Tasks/task_base.py
Thomas-McKanna: Definitely. I like perform(). It will be easy to swap out the names.

Tasks/task_base.py
Thomas-McKanna: Ok - makes sense.

Drone/drone.py
```diff
@@ -0,0 +1,18 @@
...
+from drone_base import DroneBase
+from ..Utilities import constants as c
+
+class Drone(DroneBase):
+    """
+    Concrete implementation of DroneBase. See drone__base.py for
+    documentation.
+    """
+
+    def loadDevices(self):
```
christopher-o-toole: should be load_devices. see https://www.python.org/dev/peps/pep-0008/ for official style guide
```suggestion
    def load_devices(self):
```

Drone/drone_base.py
```diff
@@ -0,0 +1,361 @@
...
+
+    cnt = 0
+
+    def __init__(self):
+        self.vehicle = None
+        self.devices = []
+        self.connected = False
+        self.logger = logging.getLogger(__name__)
+
+    def connect(self, isInSimulator):
```
christopher-o-toole:
```suggestion
    def connect(self, is_in_simulator, connection_str=c.CONNECTION_STRING_REAL,timeout=c.CONNECT_TIMEOUT):
```
and rename all corresponding instances below

Drone/drone_base.py
```diff
@@ -0,0 +1,361 @@
...
+
+        Postcondition:
+        ----------
+        Upon a successful connection, self.connected is set to True.
+        The drone is now suitable to be armed.
+
+        Returns:
+        ----------
+        None
+        """
```
christopher-o-toole: The connection string should be configurable for future implementations. See http://python.dronekit.io/guide/connecting_vehicle.html. The relevant quote is:

>You can control multiple vehicles from within a single script by calling connect() for each vehicle with the appropriate connection strings.

For future-proofing this code a bit, I'd recommend having a list or map of the strings for each drone and retrieving the connection string by drone id perhaps.

christopher-o-toole: Also, the default timeout on connect is 30 seconds, consider making a constant for this timeout. See http://python.dronekit.io/automodule.html#dronekit.connect  and ```heartbeat_timeout```

Drone/drone_base.py
```diff
@@ -0,0 +1,361 @@
...
+        if isInSimulator:
```
christopher-o-toole: The code up until line 68 does not follow DRY (don't repeat yourself), also you should be able to override the connection  value.  Consider using the following code:

```Python
if is_in_simulator:
    connection_str = c.CONNECTION_STRING_SIMULATOR
self.vehicle = dronekit.connect(connection_str, wait_ready=True, hearbeat_timeout=timeout, status_printer=None)
thread_name = threading.current_thread().name 
self.logger.info('{} : Connecting to the {} ardupilot'.format(thread_name, 'simulated' if is_in_simulator else 'real life')
```

christopher-o-toole: Note the use of the ```format``` function here, which is much more efficient and configurable than using the + operator to concatenate strings

Drone/drone_base.py
```diff
@@ -0,0 +1,361 @@
...
+        the drone's barometer
+
+        Returns:
+        ----------
+        double
+            The distance from the ground
+        """
+        return self.vehicle.location.global_relative_frame.alt
+
+    def arm(self, timeout = 60):
```
christopher-o-toole: Consider creating a constant for the arm timeout like so:
```suggestion
    def arm(self, timeout = c.DEFAULT_ARM_TIMEOUT):
```

Drone/drone_base.py
christopher-o-toole: Note that it's fine if the map or list mentioned earlier only contains one entry at the moment, it will make it easier in the long run to support adding another drone when we're ready.

Drone/drone_base.py
christopher-o-toole: Also, consider adding a mode parameter to arm the drone in a different mode (good for if we want to add our own flight mode or if we just decide to use a different flight mode)
```suggestion
    def arm(self, timeout = c.DEFAULT_ARM_TIMEOUT, mode = c.GUIDED):
```

Drone/drone_base.py
```diff
@@ -0,0 +1,361 @@
...
+        The drone is now suitable to take off.
+
+        Returns:
+        ----------
+        None
+        """
+        start = time.time()
+        self.vehicle.mode = VehicleMode(c.GUIDED_MODE)
+
+        self.logger.info(threading.current_thread().name + ": Arming")
```
christopher-o-toole: Again, DRY is not being followed. While the thread name should be retrieved on each function invocation, you can make a variable or method to retrieve it. Consider making the following utility function

```Python
def current_thread_name():
    return threading.current_thread().name
```

Drone/drone_base.py
```diff
@@ -0,0 +1,361 @@
...
+        start = time.time()
```
christopher-o-toole: Consider making a timer class for this purpose if this pattern is common throughout your code. Here's an example of one that I've created:

```Python
from timeit import default_timer as timer

class Timer():
    def __init__(self):
        self.reset()

    @property
    def elapsed(self):
        return timer() - self._start

    def reset(self):
        self._start = timer()
```

christopher-o-toole: Also, note the documentation for time is:

> Return the time in seconds since the epoch as a floating point number. Note that even though the time is always returned as a floating point number, not all systems provide time with a better precision than 1 second. While this function normally returns non-decreasing values, it can return a lower value than a previous call if the system clock has been set back between the two calls.

So this code would be vulnerable to an attack in which the system clock is changed.

Drone/drone_base.py
```diff
@@ -0,0 +1,361 @@
...
+        while (not self.vehicle.armed) and (time.time() - start < timeout):
+            self.vehicle.armed = True
+            time.sleep(1)
```
christopher-o-toole: Consider
```suggestion
            time.sleep(c.ARM_RETRY_DELAY)
```

Drone/drone_base.py
```diff
@@ -0,0 +1,361 @@
...
+        if not self.vehicle.armed:
```
christopher-o-toole: Should be replaced with regards to the DRY principle:

```Python
status_msg = 'Failed to arm' if not self.vehicle.armed else 'Armed'
logging_function = self.logger.info if self.vehicle.armed else self.logger.error
logging_function('{}: {}'.format(current_thread_name(), status_msg)

File: Drone/drone_base.py [hunk 12]
```diff
@@ -0,0 +1,361 @@
...
+        None
+
+        Returns:
+        ----------
+        Boolean
+            True if the drone is armed and false otherwise
+        """
+        return self.vehicle.armed
+
+    def takeoff(self, target_altitude, stop_event):
```
christopher-o-toole:
```suggestion
    def takeoff(self, target_altitude, stop_event, cutoff_time=c.DEFAULT_TAKEOFF_CUTOFF_TIME):
```

Drone/drone_base.py
```diff
@@ -0,0 +1,361 @@
...
+        Returns:
+        ----------
+        None
+        """
+        def takeoff_thread(target_altitude, finished, stop_event):
+            self.logger.info(threading.current_thread().name + ": Starting takeoff")
+            thrust = c.DEFAULT_TAKEOFF_THRUST
+
+            start_time = time.time()
+            cutoff_time = 10
```
christopher-o-toole:
```suggestion
            cutoff_time = timeout
```

Drone/drone_base.py
```diff
@@ -0,0 +1,361 @@
...
+
+        Postcondition:
+        ----------
+        Upon success: the drone is the requested altitude in the air. 
+
+        Returns:
+        ----------
+        None
+        """
+        def takeoff_thread(target_altitude, finished, stop_event):
```
christopher-o-toole:
```suggestion
        def takeoff_thread(target_altitude, finished, stop_event, timeout=cutoff_time):
```

Drone/drone_base.py
```diff
@@ -0,0 +1,361 @@
...
+            self.logger.info(threading.current_thread().name + ": Starting takeoff")
+            thrust = c.DEFAULT_TAKEOFF_THRUST
+
+            start_time = time.time()
+            cutoff_time = 10
+            
+            while time.time() - start_time < cutoff_time:
```
christopher-o-toole: Again, consider using a timer class and use the string.format function instead of the + operator

Drone/drone_base.py
```diff
@@ -0,0 +1,361 @@
...
+                if stop_event.is_set():
+                    self.logger.info(threading.current_thread().name + ": Takeoff halting")
+                    break
+
+                current_altitude = self.altitude()
+
+                if current_altitude >= target_altitude*0.95: # Trigger just below target alt.
```
christopher-o-toole:
```suggestion
                if current_altitude >= target_altitude*c.PERCENT_TARGET_ALTITUDE: # Trigger just below target alt.
```

Drone/drone_base.py
```diff
@@ -0,0 +1,361 @@
...
+                    break
+                elif current_altitude >= target_altitude*0.6:
```
christopher-o-toole:
```suggestion
                elif current_altitude >= target_altitude*c.PERCENT_ALTITUDE_THRUST_ADJUSTMENT_THRESHOLD:
```

Drone/drone_base.py
```diff
@@ -0,0 +1,361 @@
...
+                    thrust = c.SMOOTH_TAKEOFF_THRUST
+
+                dkw.set_attitude(self.vehicle, thrust=thrust)
+                time.sleep(1)
```
christopher-o-toole: This sleep is probably a wee bit too long for this function. Our drone is certainly powerful enough to accelerate far past the expected takeoff height in under a second.

Drone/drone_base.py
```diff
@@ -0,0 +1,361 @@
...
+            else:
+                self.logger.error("Could not take off!")
+            
+            self.logger.info(threading.current_thread().name + ": Finished takeoff")
```
christopher-o-toole: Use string.format

Drone/drone_base.py
```diff
@@ -0,0 +1,361 @@
...
+        ----------
+        The drone is on the groun (and ideally not smashed into 1000 pieces)
+        
+        Returns:
+        ----------
+        None
+        """
+        def land_thread(finished):
+            self.logger.info(threading.current_thread().name + ": Starting land")
+            while not self.vehicle.mode == VehicleMode(c.LAND_MODE):
```
christopher-o-toole: Code should follow the DRY principle. Set an intermediate variable and continue on as before 
```suggestion
            land_mode = VehicleMode(c.LAND_MODE)
            while not self.vehicle.mode == land_mode:
```

Drone/drone_base.py
```diff
@@ -0,0 +1,361 @@
...
+                self.vehicle.mode = VehicleMode(c.LAND_MODE)
```
christopher-o-toole:
```suggestion
                self.vehicle.mode = VehicleMode(land_mode)
```

Drone/drone_base.py
```diff
@@ -0,0 +1,361 @@
...
+            while self.vehicle.armed:
+                pass
+
+            self.logger.info(threading.current_thread().name + ": Finished  land")
```
christopher-o-toole: Use string.format

Drone/drone_base.py
```diff
@@ -0,0 +1,361 @@
...
+            finished.set()
+
+        finished = threading.Event()
+        threading.Thread(target=land_thread, name="LandThread-" + str(DroneBase.cnt),
```
christopher-o-toole: Use string.format

Drone/drone_base.py
```diff
@@ -0,0 +1,361 @@
...
+        ----------
+        The drone has moved in the specified direction by the specified number of meters
+
+        Returns:
+        ----------
+        threading.Event
+        """
+
+        # Calculate duration to send velocity command based on distance and velocity
+        duration = int(distance / velocity)
```
christopher-o-toole: Why is duration an int here? If I wanted to go .99 meters at 1 m/s, then I would get a duration of 0!

Drone/drone_base.py
```diff
@@ -0,0 +1,361 @@
...
+            self.logger.info(threading.current_thread().name + ": Finished  land")
+            finished.set()
+
+        finished = threading.Event()
+        threading.Thread(target=land_thread, name="LandThread-" + str(DroneBase.cnt), 
+            args=(finished,)).start()
+        DroneBase.cnt += 1
+        return finished
+
+    def move(self, direction, distance, stop_event, velocity=c.DEFAULT_VELOCITY):
```
christopher-o-toole: `velocity` is both direction and speed! You should be either using a velocity parameter only or direction and speed.

Drone/drone_base.py
```diff
@@ -0,0 +1,361 @@
...
+        # Multiply unit vector in direction by the velocity
+        vector = tuple(velocity * n for n in direction)
+
+        # Make the mavlink message
+        msg = dkw.get_velocity_message(self.vehicle.message_factory, vector)
+
+        def move_thread(msg, duration, finished, stop_event):
+            self.logger.info(threading.current_thread().name + ": Starting move")
+            # Send the message once every second
+            for _ in range(0, duration):
```
christopher-o-toole: This abuses range quit a bit, here's an example of a better timed event system:

```Python
class Timer():
    def __init__(self):
        self._event_threads = []
        self._stop_event_map = {}
        self._thread_map = {}
        self._lock = threading.Lock()
        self._num_threads_lock = threading.Lock()

        self.reset()

    def _create_event_thread(self, closure, name):
        stop_event = threading.Event()
        self._stop_event_map[name] = stop_event
        thread = threading.Thread(target=closure, daemon=True, args=(stop_event,))
        self._thread_map[name] = thread
        self._event_threads.append(thread)
        thread.start()

    @property
    def num_threads(self):
        with self._num_threads_lock as lock:
            return len(self._event_threads)

    def add_callback(self, name, when_to_call, callback, recurring=False):
        # when_to_call: call when x seconds are elapsed
        if name in self._stop_event_map:
            raise ValueError('The `name` parameter cannot be assigned the duplicate name: "{}"!'.format(name))

        if when_to_call <= self.elapsed:
            callback()
            return

        def handle_event(stop_event):
            while recurring and not stop_event.is_set():
                time.sleep(when_to_call-self.elapsed)
                event_handled = False
                while ((not event_handled) or recurring) and not stop_event.is_set():
                    if when_to_call <= self.elapsed:
                        callback()
                        event_handled = True
                        break

                    time.sleep(1e-3)

                if self.num_threads == 1 and recurring:
                    self.reset()

        self._create_event_thread(handle_event, name)

    def stop_callback(self, name):
        self._stop_event_map[name].set()

    def shutdown(self):
        for name, stop_event in self._stop_event_map.items():
            stop_event.set()
            self._thread_map[name].join()

    @property
    def elapsed(self):
        with self._lock as lock:
            return timer() - self._start

    def reset(self):
        with self._lock as lock:
            self._start = timer()
```

christopher-o-toole: This is taken from a project I'm doing rapid prototyping in at the moment, so the code above is not perfect (or well-documented for that matter), but should get the idea of how useful a object-oriented approach to timers can be

christopher-o-toole: Also, be sure to use string.format in remaining code for this file

Drone/drone_base.py
```diff
@@ -0,0 +1,361 @@
...
+        ----------
+        None
+        """
+        # Make the mavlink message
+        msg = dkw.get_velocity_message(self.vehicle.message_factory, (0, 0, 0))
+
+        def hover_thread(msg, duration, finished, stop_event):
+            self.logger.info(threading.current_thread().name + ": Starting hover")
+            # Send the message once every second
+            for _ in range(0, duration):
```
christopher-o-toole: This code is repeated again from the move_thread function and violates DRY, consider using a Timer class a shown above to schedule events.

Drone/drone_base.py
```diff
@@ -0,0 +1,361 @@
...
+            finished.set()
+
+        finished = threading.Event()
+        threading.Thread(target=hover_thread, name="HoverThread-" + str(DroneBase.cnt), 
+            args=(msg, duration, finished, stop_event)).start()
+        DroneBase.cnt += 1
+        return finished
+
+    @abc.abstractmethod
+    def loadDevices(self):
```
christopher-o-toole:
```suggestion
    def load_devices(self):
```

Drone/drone_base.py
```diff
@@ -0,0 +1,361 @@
...
+            self.logger.info(threading.current_thread().name + ": Starting hover")
+            # Send the message once every second
+            for _ in range(0, duration):
+                self.vehicle.send_mavlink(msg)
+                time.sleep(1)
+                if stop_event.is_set():
+                    self.logger.info(threading.current_thread().name + ": Hover halting")
+                    break
+
+            self.logger.info(threading.current_thread().name + ": Finished hover")
```
christopher-o-toole: Use a utility function to get the thread name

Drone/drone_base.py
```diff
@@ -0,0 +1,361 @@
...
+        msg = dkw.get_velocity_message(self.vehicle.message_factory, (0, 0, 0))
+
+        def hover_thread(msg, duration, finished, stop_event):
+            self.logger.info(threading.current_thread().name + ": Starting hover")
+            # Send the message once every second
+            for _ in range(0, duration):
+                self.vehicle.send_mavlink(msg)
+                time.sleep(1)
+                if stop_event.is_set():
+                    self.logger.info(threading.current_thread().name + ": Hover halting")
```
christopher-o-toole: Use a utility function to get the thread name

Drone/drone_base.py
```diff
@@ -0,0 +1,361 @@
...
+        ----------
+        Upon successful arm, self.vehicle.armed is True, and false otherwise.
+        The drone is now suitable to take off.
+
+        Returns:
+        ----------
+        None
+        """
+        start = time.time()
+        self.vehicle.mode = VehicleMode(c.GUIDED_MODE)
```
christopher-o-toole:
```suggestion
        self.vehicle.mode = VehicleMode(mode)
```

Drone/drone_base.py
christopher-o-toole: Also, sleeps should be configurable and be set to some default value for convenience

Drone/drone_base.py
```diff
@@ -0,0 +1,361 @@
...
+        """
+        # Make the mavlink message
+        msg = dkw.get_velocity_message(self.vehicle.message_factory, (0, 0, 0))
+
+        def hover_thread(msg, duration, finished, stop_event):
+            self.logger.info(threading.current_thread().name + ": Starting hover")
+            # Send the message once every second
+            for _ in range(0, duration):
+                self.vehicle.send_mavlink(msg)
+                time.sleep(1)
```
christopher-o-toole: Sleeps should be configurable and 1 second is likely too long.

Instructions/instruction_base.py
```diff
@@ -0,0 +1,41 @@
+# Standard Library
+import abc
+
+class InstructionBase():
+    """
+    Responsible for decoding data sent over the network and
+    using that data to create tasks.
+
+    NOTE: This class will be import once we start sending
```
christopher-o-toole:
```suggestion
    NOTE: This class will be imported once we start sending
```

Instructions/movement_instruction.py
```diff
@@ -0,0 +1,32 @@
...
+from ..Utilities import constants as c
+
+class MovementInstruction(InstructionBase):
+
+    def __init__(self, data):
+        super(MovementInstruction, self).__init__()
+        # The data here will eventually be the raw binary data that
+        # has traveled over the network. The decoding may be more
+        # involved than what is shown here.
+        (self.x, self.y, self.z) = data
```
christopher-o-toole: I believe the parentheses are not necessary here
```suggestion
        self.x, self.y, self.z = data
```

Instructions/movement_instruction.py
```diff
@@ -0,0 +1,32 @@
...
+        if self.x != 0:
+            queue.append((c.BACKWARD if self.x > 0 else c.FORWARD, abs(self.x)))
+
+        if self.y != 0:
+            queue.append((c.LEFT if self.y > 0 else c.RIGHT, abs(self.y)))
+
+        if self.z != 0:
+            queue.append((c.UP if self.z > 0 else c.DOWN, abs(self.z)))
+
+        self.movementTask = MovementTask(drone, queue)
```
christopher-o-toole:
```suggestion
        self.movement_task = MovementTask(drone, queue)
```

Instructions/movement_instruction.py
```diff
@@ -0,0 +1,32 @@
...
+
+        return self.movementTask
```
christopher-o-toole:
```suggestion
        return self.movement_task
```

Make sure to follow Python formatting standards

Instructions/movement_instruction.py
```diff
@@ -0,0 +1,32 @@
...
+        # The data here will eventually be the raw binary data that
+        # has traveled over the network. The decoding may be more
+        # involved than what is shown here.
+        (self.x, self.y, self.z) = data
+        
+        self.movementTask = None
+
+    def get_task(self, drone):
+        queue = deque()
+        if self.x != 0:
```
christopher-o-toole: Not sure how I feel about this function, it would be fairly easy to just send movement as a vector here rather than encoding each directional axis separately

Instructions/movement_instruction.py
```diff
@@ -0,0 +1,32 @@
...
+            queue.append((c.BACKWARD if self.x > 0 else c.FORWARD, abs(self.x)))
```
christopher-o-toole: Does not follow DRY principle. Consider using code analogous to the following:

```Python
direction_map = {
    'x':  (c.FORWAD, c.BACKWARD),
    'y': (c.LEFT, c.RIGHT),
    'z': (c.UP, c.DOWN)
}
positive_vector_index = 0
negative_vector_index = 1
for direction, vectors in direction_map.items():
    value = getattr(self, direction)
    if value != 0:
        queue.append((vectors[postive_vector_index if value > 0 else negative_value_index], abs(value)))
```

Utilities/constants.py
```diff
@@ -0,0 +1,34 @@
...
+DEFAULT_VELOCITY = 0.50
+VELOCITY_THRESHOLD = 2  # never let the drone go faster than 1 m/s for safety (is this a good number?)
+MINIMUM_ALLOWED_ALTITUDE = 0.5
+MAXIMUM_ALLOWED_ALTITUDE = 4.0
+RANGEFINDER_MIN = 0.29
+RANGEFINDER_EPSILON = 0.03
+
+# Connection-oriented
+CONNECTION_STRING_SIMULATOR = "tcp:127.0.0.1:5762"
+CONNECTION_STRING_REAL = "/dev/serial/by-id/usb-3D_Robotics_PX4_FMU_v2.x_0-if00"
```
christopher-o-toole: You should cite a source for this constant

Utilities/constants.py
```diff
@@ -0,0 +1,34 @@
+"""
+All of the constant values used across all of the scripts are located here.
+"""
+
+# Safety
+DEFAULT_VELOCITY = 0.50
+VELOCITY_THRESHOLD = 2  # never let the drone go faster than 1 m/s for safety (is this a good number?)
```
christopher-o-toole: Velocity is a vector, not a scalar value. You're thinking of speed here.


```suggestion
SPEED_THRESHOLD = 2  # never let the drone go faster than 1 m/s for safety (is this a good number?)
```

Utilities/drone_exceptions.py
```diff
@@ -0,0 +1,71 @@
...
+class VelocityException(Exception):
+    """
+    Too low or too high velocity
+    """
+    pass
+
+
+class BadArgumentException(Exception):
+    """
+    Caused by wrong call to function or something
```
christopher-o-toole: Not very professional to add `or something` here. Also, this should be a bit more explicit about its purpose rather than being the same as `EmergencyLandException`. Not only that, the Python standard exception for an invalid argument value is `ValueError`, so consider using that in place of your own class.

Utilities/drone_exceptions.py
```diff
@@ -0,0 +1,71 @@
...
+    """
+    pass
+
+class EmergencyLandException(Exception):
+    """
+    Caused by wrong call to function or something
```
christopher-o-toole: Was this intentional? It seems like this description should be akin to:

Thrown when an unrecoverable state is reached and the drone should land for the preservation of its safety and the safety of others.

Utilities/dronekit_wrappers.py
```diff
@@ -0,0 +1,93 @@
...
+            velocity_y, # Y velocity in NED frame in m/s
+            velocity_z, # Z velocity in NED frame in m/s
+            0, 0, 0, # afx, afy, afz acceleration (not supported yet, ignored in GCS_Mavlink)
+            0, 0)    # yaw, yaw_rate (not supported yet, ignored in GCS_Mavlink)
+   
+
+def set_attitude(vehicle, roll_angle = 0.0, pitch_angle = 0.0, yaw_rate = 0.0, thrust = 0.5):
+    """
+    NOTE: CURRENTLY NOT USED
+    The follow comments are from the DroneKit website:
```
christopher-o-toole:
```suggestion
    The following comments are from the DroneKit website:
```

Utilities/dronekit_wrappers.py
```diff
@@ -0,0 +1,93 @@
...
+
+    Returns:
+    ----------
+    MAVLink_message (a DroneKit object)
+    """
+    return message_factory.set_position_target_global_int_encode(
+            0,       # time_boot_ms (not used)
+            0, 0,    # target system, target component
+            mavutil.mavlink.MAV_FRAME_GLOBAL_RELATIVE_ALT_INT, # frame
+            0b0000111111000111, # type_mask (only speeds enabled)
```
christopher-o-toole: Consider creating an enum for these bitmasks

Utilities/constants.py
```diff
@@ -0,0 +1,34 @@
...
+# Connection-oriented
+CONNECTION_STRING_SIMULATOR = "tcp:127.0.0.1:5762"
+CONNECTION_STRING_REAL = "/dev/serial/by-id/usb-3D_Robotics_PX4_FMU_v2.x_0-if00"
+
+# DroneKit Vehicle Modes
+GUIDED_MODE = "GUIDED"
+LAND_MODE = "LAND"
+
+# Directions
+UP = (0, 0, 1)
```
christopher-o-toole: This is the perfect case for an enumeration like the following:

```
from enum import Enum
class Directions(Enum):
    UP = (0, 0, 1)
    # ...
```

Note that you will have to use the `Directions.UP.value` format to get the underlying enumeration value.

Utilities/dronekit_wrappers.py
christopher-o-toole: Also you should probably link the URL as well instead of simply stating that they're from the website

Utilities/lock.py
```diff
@@ -0,0 +1,25 @@
...
+    lock: threading.Lock
+        A static variable that is shared among all threads
+    """
+    lock = None
+
+    def __init__(self):
+        self.lock = None
+
+    @staticmethod
+    def getLock():
```
christopher-o-toole:
```suggestion
    def get_lock():
```

Drone/drone_base.py
```diff
@@ -0,0 +1,361 @@
...
+import dronekit
+from dronekit import VehicleMode
+import logging
+import math
+import time
+import threading
+
+# Ours
+from ..Utilities import constants as c
+from ..Utilities import dronekit_wrappers as dkw
```
christopher-o-toole: Consider using a slightly more descriptive name than dkw, it could be confusing to developers whom are not as familiar with the code.

christopher-o-toole: Either that or establish the standard of always importing the module by that name.

Utilities/dronekit_wrappers.py
```diff
@@ -0,0 +1,93 @@
...
+        radians(yaw_rate), # Body yaw rate in radian
+        thrust  # Thrust
+    )
+    vehicle.send_mavlink(msg)
+
+def to_quaternion(roll = 0.0, pitch = 0.0, yaw = 0.0):
+    """
+    NOTE: CURRENTLY NOT USED
+
+    Convert degrees to quaternions
```
christopher-o-toole: Cite a source for the formulas applied for quaternions here.

christopher-o-toole (changes requested):
This is a pretty good initial attempt at writing the new flight interface, but there's a few issues that need to be addressed before approval. In a nutshell:

- Don't repeat yourself (DRY principle). If you write code that is the same with the exception of a few variable values/constants, you're doing something wrong. Repeated code is hard to maintain since bug fixes must be repeated across all instances where the code appears.
- DRY also applies to repeated instantiations of classes instead of just creating it once and caching that value.
- Velocity should always be a vector, it is not the same thing as speed which is the magnitude of velocity.
- Make sure to follow PEP 8 which dictates using snake case for method and member variable names.
- `+` operator string concatenation is not preferred for performance reasons. Strings are immutable, so every time you use the + operator to concatenate two literal values you are creating two string objects, creating a new concatenated string object, and deallocating the two original string objects. Usage of buffers (i.e. ','.join('a', 'b') == 'a,b' is true) or the format function (i.e. '{} {}'.format('big', 'house') == 'big house' is true) helps alleviate this problem
- Magic numbers: all constants should be configurable and preferably documented to include how they were obtained. Hardcoded values like `.95` and `.6` with no explanation is bad style.

Drone/drone.py
```diff
@@ -0,0 +1,18 @@
+import threading
+import logging
+
+# Ours
+from drone_base import DroneBase
+from ..Utilities import constants as c
```
ClayMav: c is a bad name for this variable, and a note, I don't like the usage of the relative `..` import, so in the future lets package all of this stuff into more broad directories with focuses such as "runs on drone" and "runs on base station", not sure how we will name this but I think its a better way to do this. This will allow us to import with `flight.utilities` instead. Not requiring for this commit since tools is going to get 🔫 (restructured) soon.

Drone/drone.py
```diff
@@ -0,0 +1,18 @@
...
+
+class Drone(DroneBase):
+    """
+    Concrete implementation of DroneBase. See drone__base.py for
```
ClayMav: Concrete is ambiguous, please describe with something more specific

Drone/drone.py
```diff
@@ -0,0 +1,18 @@
...
+    documentation.
+    """
+
+    def loadDevices(self):
+        pass
```
ClayMav, line 12: Since this is a `pass` function, can you add documentation for what this will eventually do since it has no purpose yet.

Drone/drone.py
```diff
@@ -0,0 +1,18 @@
...
+
+    # Other drone specific functions
```
ClayMav: This should not be here, instead make a more descriptive version of this the docstring at the top of the file.

Drone/drone_base.py
```diff
@@ -0,0 +1,361 @@
+# Standard Library
+import abc
+import coloredlogs
+import dronekit
+from dronekit import VehicleMode
```
ClayMav: Don't import both dronekit and the VehicleMode from dronekit, either get all the contents out as you did with the `from import` or just `import dronekit` and do `dronekit.VehicleMode`
This is a lot of imports, can any of these be moved to other places? Not a suggested change, just a thought you should give here, less imports is better.

Drone/drone_base.py
```diff
@@ -0,0 +1,361 @@
...
+import logging
+import math
+import time
+import threading
+
+# Ours
+from ..Utilities import constants as c
```
ClayMav, line 9: As with before, c is a bad name

Drone/drone_base.py
ClayMav: I would prefer just using the full name

Drone/drone_base.py
```diff
@@ -0,0 +1,361 @@
...
+from ..Utilities import dronekit_wrappers as dkw
+
+class DroneBase(object):
+    """
+    Wraps a DroneKit.vehicle and sensors
```
ClayMav: Does it wrap a vehicle and its associated sensors or does it wrap a vehicle and in addition to that, wrap sensors?

Drone/drone_base.py
```diff
@@ -0,0 +1,361 @@
...
+        the drone's rangefinder
+
+        Returns:
+        ----------
+        double
+            The distance from the ground
+        """
+        return self.vehicle.rangefinder.distance
+
+    def altitude_failsafe(self):
```
ClayMav: Should be a property

Drone/drone_base.py
```diff
@@ -0,0 +1,361 @@
...
+            self.logger.info(threading.current_thread().name + ": Connecting to the simulated ardupilot")
+        else:
+            self.vehicle = dronekit.connect(c.CONNECTION_STRING_REAL, wait_ready=True, 
+                status_printer=None)
+            self.logger.info(threading.current_thread().name + ": Connecting to the real-life ardupilot")
+
+        if self.vehicle is not None:
+            self.connected = True
+
+    def altitude(self):
```
ClayMav: Should be a property

Drone/drone.py
ClayMav: I changed my mind, you should be importing each constant individually 
```python3
from ..Utilities.constants import list, each, constant, here
```

Drone/drone_base.py
ClayMav: Also as with before import each constant individually

Drone/drone_base.py
ClayMav: Appending to this form my comments above, use `CONNECTION_STRING_REAL` here and not `c.CONNECTION_STRING_REAL`

Drone/drone_base.py
```diff
@@ -0,0 +1,361 @@
...
+        start = time.time()
+        while (not self.vehicle.armed) and (time.time() - start < timeout):
+            self.vehicle.armed = True
+            time.sleep(1)
+        if not self.vehicle.armed:
+            self.logger.error(threading.current_thread().name + ": Failed to arm")
+        else:
+            self.logger.info(threading.current_thread().name + ": Armed")
+
+    def is_connected(self):
```
ClayMav: Should be property

Drone/drone_base.py
```diff
@@ -0,0 +1,361 @@
...
+        None
+
+        Returns:
+        ----------
+        Boolean
+            True if the drone is connected to Ardupilot and false otherwise
+        """
+        return self.connected
+
+    def is_armed(self):
```
ClayMav: Should be property

Drone/drone_base.py
```diff
@@ -0,0 +1,361 @@
...
+        Returns:
+        ----------
+        None
+        """
+        def land_thread(finished):
+            self.logger.info(threading.current_thread().name + ": Starting land")
+            while not self.vehicle.mode == VehicleMode(c.LAND_MODE):
+                self.vehicle.mode = VehicleMode(c.LAND_MODE)
+            while self.vehicle.armed:
+                pass
```
ClayMav, line 254: Might want to make this a smaller sleep, no need to hog the thread like that

Drone/drone_base.py
```diff
@@ -0,0 +1,361 @@
...
+        finished = threading.Event()
+        threading.Thread(target=hover_thread, name="HoverThread-" + str(DroneBase.cnt), 
+            args=(msg, duration, finished, stop_event)).start()
+        DroneBase.cnt += 1
+        return finished
+
+    @abc.abstractmethod
+    def loadDevices(self):
+        """
+        Behavior of this function is currently undefined.
```
ClayMav: Delete it then if it has no purpose, or define its future purpose

Drone/drone_controller.py
```diff
@@ -0,0 +1,127 @@
...
+import sys
+
+# Ours
+from drone import Drone
+from drone_controller_base import DroneControllerBase
+from ..Instructions.movement_instruction import MovementInstruction
+from ..Tasks.hover_task import HoverTask
+from ..Tasks.movement_task import MovementTask
+from ..Tasks.takeoff_task import TakeoffTask
+from ..Utilities import constants as c
```
ClayMav, line 14: You know the deal

Drone/drone_controller.py
```diff
@@ -0,0 +1,127 @@
...
+from ..Utilities import drone_exceptions as de
```
ClayMav: just use the real name

Drone/drone_controller.py
```diff
@@ -0,0 +1,127 @@
...
+        self.logger = logging.getLogger(__name__)
+        self.debug = False
+        coloredlogs.install(level='DEBUG')
+        # The following two lines are purely for testing purposes. Instructions
+        # will be pushed onto the heap as a result of the swarm controller
+        # sending instructions or inter-drone communication.
+        heapq.heappush(self.instruction_queue, (0, MovementInstruction((5, 5, 0))))
+        heapq.heappush(self.instruction_queue, (0, MovementInstruction((-5, -5, 0))))
+
+    def set_id(self):
```
ClayMav: should be a setter `@id.setter` though id is not there

Drone/drone_controller.py
```diff
@@ -0,0 +1,127 @@
...
+                    stop_hover_event = self.task.exit_task()
+                    stop_hover_event.wait()
+                    self.task = None
+                self.read_next_instruction()
+            # If there are no instructions, begin to hover
+            else:
+                if self.task is None:
+                    self.task = HoverTask(self.drone)
+         
+    def run_loop(self):
```
ClayMav, line 44: This should be begin or something since it is basically the start, it should also be implemented such that it would be run using a `with` statement

Drone/drone_controller_base.py
```diff
@@ -0,0 +1,104 @@
...
+        ----------
+        The instruction queue contains one less element.
+
+        Returns:
+        ----------
+        None
+        """
+        pass
+
+    @abc.abstractmethod
```
ClayMav, line 51: setter

ClayMav (changes requested):
Nothing severe in here, but get these all cleaned up and we can get it merged.

Drone/drone_base.py
Thomas-McKanna: Would it be appropriate for the map to be defined in the constants file or in drone_base.py?

Drone/drone_base.py
Thomas-McKanna: I have changed all instances of using + for concatenation to using the format function.

Drone/drone_base.py
Thomas-McKanna: I have added a new file called helper.py in the Utilities folder and put the current_thread_name() in it.

Drone/drone_base.py
Thomas-McKanna: I have defined the suggested Timer class in the Utilities/timer.py.

Drone/drone_base.py
Thomas-McKanna: I have also added the constant c.MAV_MESSAGE_RESEND_DELAY to fix other similar issues.

Drone/drone_base.py
Thomas-McKanna: As per your previous suggestion, I gave takeoff_thread() an additional parameter for cutoff time with a default value. So I just got rid of the declaration of cutoff_time altogether since it is now an argument.

Drone/drone_base.py
Thomas-McKanna: What do you think would be a good default value for the sleep? I have set default to 0.25 for now.

Drone/drone_base.py
Thomas-McKanna: What is the suggestion here? Should I make a local variable named land_mode?

Drone/drone_base.py
Thomas-McKanna: The reason is that the mavlink message is sent every second (see the note at http://python.dronekit.io/examples/guided-set-speed-yaw-demo.html#send-global-velocity). The value is being used like this:
`for _ in range(0, duration)`
around drone_base.py line 309 so a float would cause error.

I don't know if sending every second is a requirement or not - we can do some experimenting.

Drone/drone_base.py
Thomas-McKanna: c.DEFAULT_VELOCITY has been changed to c.DEFAULT_SPEED and the parameter velocity in move has been renamed to speed.

Drone/drone_base.py
Thomas-McKanna: Nice code! I have swapped out the old range based implementation with your timer class. I also changed the hover function, which was very similar to move().

Drone/drone_base.py
Thomas-McKanna: I believe I have fixed the DRY problems with hover() and with move(). How does this look:
```python
    def hover(self, duration, stop_event,
        send_delay=c.MAV_VELOCITY_MSG_RESEND_DELAY):
        # Make the mavlink message
        msg = dkw.get_velocity_message(self.vehicle.message_factory, (0, 0, 0))

        def hover_thread(msg, duration, finished, stop_event,
            delay=send_delay):
            self.logger.info('{}: Starting hover'.format(current_thread_name()))
            # Send the message once every delay seconds
            timer = Timer()
            timer.add_callback(
                "move_msg", delay, 
                lambda : self.vehicle.send_mavlink(msg),
                recurring=True)

            if stop_event.wait(duration):
                msg = 'Hover halting'
            else:
                msg = 'Hover finished'
            
            timer.stop_callback("move_msg")
            self.logger.info(
                '{}: {}'.format(current_thread_name(), msg))
            
            finished.set()

        finished = threading.Event()
        thread_name = 'HoverThread-{}'.format(str(DroneBase.cnt))
        threading.Thread(
            target=hover_thread, name=thread_name, 
            args=(msg, duration, finished, stop_event)
            ).start()
        DroneBase.cnt += 1
        return finished
```

Drone/drone_base.py
Thomas-McKanna: I have changed it so that mode is a parameter of arm and defaults to c.GUIDED. Is this what you were suggesting here?

Drone/drone_base.py
Thomas-McKanna: The sleep duration is now configurable by setting c.MAV_HOVER_MSG_RESEND_DELAY. I still have it defaulting to 1 sec, as is suggested on the dronekit website, but its value can easily be changed.

Instructions/movement_instruction.py
Thomas-McKanna: The code runs fine without them, so they have been removed.

Instructions/movement_instruction.py
Thomas-McKanna: I don't disagree with you. For now, I will implement the suggested code for your next comment. I anticipate we will be doing a major overhaul of the lower-level details of flight soon and we can address the issue then.

Utilities/constants.py
Thomas-McKanna: I am not sure on the source for either of them. I copied the simulator one from the original flight code that existed before I started working on flight and the second one from Innocent. I tried looking online but can't find any source.

Utilities/drone_exceptions.py
Thomas-McKanna: I overlooked the docstring of BadArgumentException. I have removed the exception entirely. The EmergencyLandException docstring must have been copied from BadArgumentException and never changed. I have replaced its docstring with an appropriate one.

Utilities/drone_exceptions.py
Thomas-McKanna: It was not and I have changed all of the Exception docstrings to be a bit more descriptive and uniform.

Utilities/dronekit_wrappers.py
Thomas-McKanna: It seems I have erroneously put the comment under set_attitude() when it should be with send_global velocity(). I have changed its position and provided a link to the relevant web page.

Utilities/dronekit_wrappers.py
Thomas-McKanna: Could you elaborate on this suggestion more? I'm not sure what you mean.

Drone/drone_base.py
christopher-o-toole: Yes, it is preferred to create local variables which are assigned to the constant in order to avoid repetitive referencing of the module constant name

Drone/drone_base.py
christopher-o-toole: Yes.

Drone/drone_base.py
Thomas-McKanna: I have removed the cast to integer. Duration no longer needs to be an integer because the Timer class is being used, so the range-based for loop is no longer needed.

Drone/drone_base.py
Thomas-McKanna: I have changed it so that it uses the full name.

Drone/drone.py
Thomas-McKanna: I started to make the suggested changes to drone_base.py, but things quickly became bloated. Here is what it would look like:
```
from ..Utilities.constants import CONNECTION_STRING_REAL, CONNECTION_STRING_SIMULATOR, 
HEARTBEAT_TIMEOUT, DEFAULT_ARM_TIMEOUT, GUIDED, ARM_RETRY_DELAY, DEFAULT_TAKEOFF_CUTOFF_TIME, 
DEFAULT_TAKEOFF_THRUST, PERCENT_TARGET_ALTITUDE, PERCENT_ALTITUDE_THRUST_ADJUSTMENT_THRESHOLD,
SMOOTH_TAKEOFF_THRUST, MAV_ATTITUDE_MSG_RESEND_DELAY, LAND, MAV_VELOCITY_MSG_RESEND_DELAY,
MAV_HOVER_MSG_RESEND_DELAY
```
One thing I am noticing is that many of the constants are only used in one file. Would it be a good idea to define file-specific constants at the top of each file and only put shared constants in the constants folder.

Or we could import constants (not as c) and have write out constants.SOME_VALUE for all of them.

Thoughts?

Drone/drone.py
Thomas-McKanna: I have adjusted the docstring to read as follows:
```python
"""
A drone that can perform various tasks required for Mission 8. See
DroneBase for documentation. Add custom drone behavior to this file.
"""
```

Is this still too vague?

Drone/drone.py
Thomas-McKanna: I removed the comment and rephrased its meaning in the docstring at the top of the file.

Drone/drone_base.py
Thomas-McKanna: I have removed the import of VehicleMode. 

Unless there are some tricks I don't know about, the only way to cut down on imports would be to merge some files together. I will keep this in mind as the project continues and if there is an opportunity to do so, I will put it in another pull request.

Drone/drone.py
Thomas-McKanna: I have added the following docstring:
```python
"""
Load a list of interfaces to physical devices, such as sensors.

After calling this function, self.devices will be ready to use.
"""
```

Drone/drone_controller.py
Thomas-McKanna: I moved run_loop() up to below __init__().

In what way do you suggest the use of a `with` statement?

Drone/drone_controller_base.py
Thomas-McKanna: What is the suggestion?

Utilities/dronekit_wrappers.py
Thomas-McKanna: This is another one that was around before I joined flight. I was able to find an explanation on wikipedia at https://en.wikipedia.org/wiki/Conversion_between_quaternions_and_Euler_angles#Source_Code and will put this link in the code.

Thomas-McKanna: This was one of the functions that was around before flight. I was able to find an explanation on wikipedia: https://en.wikipedia.org/wiki/Conversion_between_quaternions_and_Euler_angles#Source_Code

Drone/drone_base.py
christopher-o-toole: Looks alright to me.

Utilities/constants.py
christopher-o-toole: Ask Innocent, he may know.

Utilities/dronekit_wrappers.py
Thomas-McKanna: I have made an enum for each bitmask (send_velocity and set_attitude).

Drone/drone.py
ClayMav: I appreciate your example, I am incorrect, do as you were. it feels wrong to me but its clearly the way to be. However, look for lots of examples where enums could be used. Use constants that have use in multiple files in the constants.py file, but if it is only going to be relevant in one file, don't bloat constants.

Drone/drone_controller.py
ClayMav: When the object is instanciated, is there any reason why `run_loop()` should not be run?

Drone/drone_controller_base.py
ClayMav: Since the function is called `set_id`, it should be an `@setter`

Drone/drone_base.py
Thomas-McKanna: I have rephrased the docstring so that it is clearer:
```python
"""
Interface to dronekit.vehicle and to sensors.
"""
```

Drone/drone_base.py
Thomas-McKanna: I have turned all sleep values into constants which can be configured from the constants file.

Drone/drone_base.py
Thomas-McKanna: I have replaced the docstring with what I believe the purpose of the function will be:
```python
"""
Load a list of interfaces to physical devices, such as sensors.

After calling this function, self.devices will be ready to use.
"""
```

Drone/drone_controller.py
Thomas-McKanna: It should always be called, otherwise object does nothing. I have not used `with` statements much so I'm not sure how it should be used in this case. Could you give an example of what you were thinking?

Utilities/constants.py
Thomas-McKanna: I was able to get sources from Innocent and I have commented them in the code.

Drone/drone_controller_base.py
Thomas-McKanna: I have removed set_id as an abstract function (since it was trivial) and made id a property of DroneControllerBase. I also added a getter, which didn't exist before. Here is the result:

```python
    @property
    def id(self):
        """
        Get the drone's id
        """
        return self._id

    @id.setter
    def id(self, identifier):
        """
        Set the drone's id.
        """
        self._id = identifier
```

Drone/drone_controller.py
ClayMav: basically a `with` statement will, when used, create an instance of the object, `DroneController` in this case, and then run some functions on it.
It basically encapsulates the creation and deletion of an object. You can choose whether or not you feel it is necessary.
However, for `DroneController`, the class may as well just run the function `run_loop` when the object is created (in `__init__`), since as you said, if the function is not called, the object does literally nothing.

Drone/drone.py
ClayMav: That is fine, the previous description was not a description of the class, the first line of this new one is a good description of what the class is going to do.

Thomas-McKanna:
I've made some changes in addition to what you all have suggested:

- Variables and functions that were only meant for internal use of a class have had a '_' added to the front of their names.
- All getters and setters are not implemented via @property.
- I've gone through all docstrings and made sure they follow the numpy docstring format.
- I have gone through and made all the code fit within 79 columns (https://www.python.org/dev/peps/pep-0008/#maximum-line-length).
- I have fixed all the Codacy errors for all non-tool code.
- Flight modes are now an enum.

ClayMav:
Might want to push those changes

Drone/drone_base.py
Thomas-McKanna: I have put the map in the constants folder. Here is what it looks like:
```python
DRONES = {
        'Leonardo_Sim' : 'tcp:127.0.0.1:5762',
        'Leonardo': '/dev/serial/by-id/usb-3D_Robotics_PX4_FMU_v2.x_0-if00'
        }
```
I figured we could name the drones based on each of the turtles, and append "Sim" to the name if it is the simulated version.

christopher-o-toole (approved):

ClayMav (approved):

PR #13, opened 2019-01-24
Add simulation multiplier variable.

Thomas-McKanna:
In Gazebo, the real-time factor (which varies from computer to computer) often causes the drone to travel different distances depending on whose computer the code is running on. This pull request introduces a SIMULATION_MULTIPLIER constant which increases the duration the drone flies in a given direction by a constant value.

Intended use case: user observes their average real-time factor (let it be x), calculates SIMULATION_MULTIPLIER by doing 1.0/x, and then sets the constant to that value.

The proposed changes worked upon testing in Gazebo.


Utilities/constants.py
```diff
@@ -54,6 +54,7 @@ class DronekitBitmasks(Enum):
 SMOOTH_TAKEOFF_THRUST = 0.6
 PERCENT_TARGET_ALTITUDE = 0.95
 PERCENT_ALTITUDE_THRUST_ADJUSTMENT_THRESHOLD = 0.6
+SIMULATION_MULTIPLIER = 1 # Divide 1 by your avergage real-time factor
```
ClayMav:
```suggestion
SIMULATION_MULTIPLIER = 1 # Divide 1 by your average real-time factor
```

ClayMav (changes requested):
One word misspelled, this looks good to go otherwise

ClayMav (approved):

christopher-o-toole (approved):
You might consider making a command-line argument for this, as the multiplier may change across systems depending on potatoness and those changes will potentially create merge conflicts (i.e. thomas pushes code with multiplier .5 but andrew has multiplier .1), but it looks fine in general.

PR #14, opened 2019-01-24
Add custom priority queue and use it in controller.

Thomas-McKanna:
A new priority queue class to abstract the details of pushing and popping the queue. The priority queue is used in drone controller to keep track of instructions. Benefits of this pull request are:

- Constant enum for three priority levels: LOW, MEDIUM, and HIGH (previously hard-coded)
- Abstraction of popping the queue, which used to require indexing into the popped item to retrieve the desired data

I wrote unit tests for this one, as I plan to do for all future pull requests.

### DIFF
--- a/Drone/drone_controller.py
+++ b/Drone/drone_controller.py
@@ -32,7 +32,7 @@ class DroneController(DroneControllerBase):
         self.drone = drone
         self._instruction_queue = PriorityQueue()
 
-        coloredlogs.install(level='DEBUG')
+        coloredlogs.install(level=logging.INFO)
 
         # The following two lines are purely for testing purposes. Instructions
         # will be pushed onto the heap as a result of the swarm controller
@@ -106,7 +106,7 @@ class DroneController(DroneControllerBase):
 
         if isinstance(self._task, HoverTask) or self._task is None:
             # Process remaining instructions
-            if self._instruction_queue.size:
+            if not self._instruction_queue.empty():
                 # Stop hovering, if we were doing so
                 if isinstance(self._task, HoverTask):
                     stop_hover_event = self._task.exit_task()
--- a/Drone/drone_controller.py
+++ b/Drone/drone_controller.py
@@ -15,6 +15,7 @@ from ..Utilities import constants as c
 from ..Utilities import drone_exceptions
 from ..Utilities.helper import current_thread_name
 from ..Utilities.lock import SharedLock
+from ..Utilities.priority_queue import PriorityQueue
 
 class DroneController(DroneControllerBase):
     """Manages the control logic of a drone.
@@ -28,18 +29,20 @@ class DroneController(DroneControllerBase):
 
         self._logger = logging.getLogger(__name__)
         self._debug = True
-
         self.drone = drone
+        self._instruction_queue = PriorityQueue()
 
         coloredlogs.install(level='DEBUG')
 
         # The following two lines are purely for testing purposes. Instructions
         # will be pushed onto the heap as a result of the swarm controller
         # sending instructions or inter-drone communication.
-        heapq.heappush(
-            self._instruction_queue, (0, MovementInstruction((5, 5, 0))))
-        heapq.heappush(
-            self._instruction_queue, (0, MovementInstruction((-5, -5, 0))))
+        sample_data1 = (5, 5, 0)
+        sample_data2 = (-5, -5, 0)
+        self._instruction_queue.push(
+            c.Priorities.MEDIUM, MovementInstruction(sample_data1))
+        self._instruction_queue.push(
+            c.Priorities.MEDIUM, MovementInstruction(sample_data2))
 
     def run_loop(self):
         SharedLock.get_lock().acquire()
@@ -103,7 +106,7 @@ class DroneController(DroneControllerBase):
 
         if isinstance(self._task, HoverTask) or self._task is None:
             # Process remaining instructions
-            if len(self._instruction_queue):
+            if self._instruction_queue.size:
                 # Stop hovering, if we were doing so
                 if isinstance(self._task, HoverTask):
                     stop_hover_event = self._task.exit_task()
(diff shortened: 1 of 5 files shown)


Drone/drone_controller.py
```diff
@@ -103,7 +106,7 @@ def _update(self):
 
         if isinstance(self._task, HoverTask) or self._task is None:
             # Process remaining instructions
-            if len(self._instruction_queue):
```
ClayMav, line 106: I would prefer if you would use the `len()` function actually
https://kushaldas.in/posts/len-function-in-python.html

Tests/priority_queue_test.py
```diff
@@ -0,0 +1,37 @@
...
+        """Test that items are able to be inserted an removed."""
+        queue = PriorityQueue()
+        sample_item = 5
+        queue.push(Priorities.MEDIUM, sample_item)
+        returned_item = queue.pop()
+        self.assertEqual(sample_item, returned_item, "Not inserting and \
+        removing items correctly.")
+
+    def test_proper_order(self):
+        """Test that a mixed-bag of priorities comes out in the right order."""
```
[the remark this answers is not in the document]
ClayMav: Get on it! Breaking bug right here!

Utilities/priority_queue.py
```diff
@@ -0,0 +1,54 @@
...
+        An item, or None if the priority queue is empty.
+        """
+        if (self.size > 0):
+            # Index 0 is priority, index 1 is the item.
+            return heapq.heappop(self._queue)[1]
+        else:
+            return None
+
+    @property
+    def size(self):
```
ClayMav: I would just leave this out

ClayMav (changes requested):
I also notice you putting the second `"""` in a docstring on the next line a lot, its not necessary unless its over the 80 character limit, in which case, go from 
```python
"""Hello World
"""
# to this:
"""
Hello World
"""
```

Drone/drone_controller.py
Thomas-McKanna: Your right, `len()` would be more readable than calling a size property. I just had an idea though. What are your thoughts on me adding an `empty()` method to the PriorityQueue object? Then the code could read `if not self._instruction_queue.empty()`, which has a really clear intent. (I will still overload `len()`, by the way)

Utilities/priority_queue.py
Thomas-McKanna: Agreed. Out it goes.

christopher-o-toole: Agreed. Overriding the `__len__` method is preferred here.

Utilities/constants.py
```diff
@@ -30,6 +29,12 @@ class Drones(Enum):
 RANGEFINDER_EPSILON = 0.03
 DEFAULT_ARM_TIMEOUT = 60
 
+# Values based on heapq, with is a min-heap.
+class Priorities(Enum):
+    LOW = 3
+    MEDIUM = 2
+    HIGH = 1
+
 # DroneKit Vehicle Modes
```
christopher-o-toole, line 45: Make this a docstring under the class, otherwise this information will not appear in typical intellisense parsers.

Utilities/constants.py
```diff
@@ -30,6 +29,12 @@ class Drones(Enum):
 RANGEFINDER_EPSILON = 0.03
 DEFAULT_ARM_TIMEOUT = 60
 
+# Values based on heapq, with is a min-heap.
```
christopher-o-toole: Again, make this a docstring instead of a comment.

Drone/drone_controller.py
```diff
@@ -28,18 +29,20 @@ def __init__(self, drone):
 
         self._logger = logging.getLogger(__name__)
         self._debug = True
-
         self.drone = drone
+        self._instruction_queue = PriorityQueue()
 
         coloredlogs.install(level='DEBUG')
```
christopher-o-toole: Consider making an enum for logging levels or using the built-in logging mode constants if they achieve the same effect here.

Tests/priority_queue_test.py
```diff
@@ -0,0 +1,37 @@
...
+        returned_item = queue.pop()
+        self.assertEqual(sample_item, returned_item, "Not inserting and \
+        removing items correctly.")
+
+    def test_proper_order(self):
+        """Test that a mixed-bag of priorities comes out in the right order.""" 
+        queue = PriorityQueue()
+        items = [1, 2, 3]
+        priorities = [Priorities.LOW, Priorities.MEDIUM, Priorities.HIGH]
+        for x in range(0, 3):
```
christopher-o-toole:
```suggestion
        for item, priority in zip(items, priorities):
```

Tests/priority_queue_test.py
```diff
@@ -0,0 +1,37 @@
...
+            queue.push(priorities[x], items[x])
```
christopher-o-toole:
```suggestion
            queue.push(priority, item)
```

Tests/priority_queue_test.py
```diff
@@ -0,0 +1,37 @@
...
+
+        items.reverse()
+        for x in range(0, 3):
```
christopher-o-toole:
```suggestion
        for item in items:
```

Tests/priority_queue_test.py
```diff
@@ -0,0 +1,37 @@
...
+            self.assertEqual(items[x], queue.pop(), "Not returning items with \
```
christopher-o-toole:
```suggestion
            self.assertEqual(item, queue.pop(), "Not returning items with \
```

Tests/priority_queue_test.py
```diff
@@ -0,0 +1,37 @@
...
+            higher priority first.")
+
+    def test_empty_queue(self):
+        """Test that popping an empty queue does not throw error."""
```
christopher-o-toole:
```suggestion
        """Test that popping an empty queue does not throw an error."""
```

christopher-o-toole (changes requested):
There are a few grammar and code issues which should be fixed before approval, but this request didn't have too many issues overall.

christopher-o-toole (approved):
Issues were resolved as requested.

ClayMav (approved):
Alright with me.

PR #15, opened 2019-01-27
Maintain order for items of equal priority.

Thomas-McKanna:
Items of equal priority inserted into a PriorityQueue did not maintain order. For instance, if two movement instruction of equal priority were pushed onto the queue, they would be executed in reverse order, which is probably not what was intended.

I'm not sure if my solution is the best way to do it, but it gets the job done. PriorityQueue keeps an internal counter starting at 987654. When items are inserted, their priority level is appended to the front of this counter. So if I insert two movements of equal priority the first gets id 3987654 and the second gets 3987655, and so the first one gets popped out first (since it's a min-heap).

This method would cease to work after 1000000 - 987654 = 12346 items were queued, after which the number would jump in magnitude and cause things to break (11000000 > 3999999).

### DIFF
--- a/Tests/priority_queue_test.py
+++ b/Tests/priority_queue_test.py
@@ -16,12 +16,13 @@ class TestPriorityQueue(unittest.TestCase):
     def test_proper_order(self):
         """Test that a mixed-bag of priorities comes out in the right order."""
         queue = PriorityQueue()
-        items = [1, 2, 3]
-        priorities = [Priorities.LOW, Priorities.MEDIUM, Priorities.HIGH]
+        items = [1, 2, 3, 4]
+        priorities = [Priorities.LOW, Priorities.LOW, Priorities.MEDIUM,
+            Priorities.HIGH]
         for item, priority in zip(items, priorities):
             queue.push(priority, item)
 
-        items.reverse()
+        items = [4, 3, 1, 2]
         for item in items:
             self.assertEqual(item, queue.pop(), "Not returning items with \
             higher priority first.")
--- a/Utilities/priority_queue.py
+++ b/Utilities/priority_queue.py
@@ -12,6 +12,8 @@ class PriorityQueue():
         The priority queue.
     """
 
+    count = 987654
+
     def __init__(self):
         self._queue = []
 
@@ -27,7 +29,9 @@ class PriorityQueue():
             The data being inserted into the queue.
 
         """
-        heapq.heappush(self._queue, (priority.value, item))
+        val = int('{}{}'.format(priority.value, PriorityQueue.count))
+        PriorityQueue.count = PriorityQueue.count + 1
+        heapq.heappush(self._queue, (val, item))
 
     def pop(self):
         """Remove the next item from the priortiy queue.
--- a/Utilities/priority_queue.py
+++ b/Utilities/priority_queue.py
@@ -1,5 +1,7 @@
 import heapq
 
+DEFAULT_COUNT = 900000
+
 class PriorityQueue():
     """Custom priority queue implementation.
 
@@ -12,7 +14,7 @@ class PriorityQueue():
         The priority queue.
     """
 
-    count = 900000
+    count = DEFAULT_COUNT
 
     def __init__(self):
         self._queue = []
@@ -30,7 +32,7 @@ class PriorityQueue():
 
         """
         val = int('{}{}'.format(priority.value, PriorityQueue.count))
-        PriorityQueue.count = PriorityQueue.count + 1
+        PriorityQueue.count += 1
         heapq.heappush(self._queue, (val, item))
 
     def pop(self):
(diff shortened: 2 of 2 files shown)


Tests/priority_queue_test.py
```diff
@@ -16,12 +16,13 @@ def test_insertion(self):
     def test_proper_order(self):
         """Test that a mixed-bag of priorities comes out in the right order."""
         queue = PriorityQueue()
-        items = [1, 2, 3]
-        priorities = [Priorities.LOW, Priorities.MEDIUM, Priorities.HIGH]
+        items = [1, 2, 3, 4]
+        priorities = [Priorities.LOW, Priorities.LOW, Priorities.MEDIUM,
```
ClayMav, line 20: theres 2 LOWs here

Tests/priority_queue_test.py
```diff
@@ -16,12 +16,13 @@ def test_insertion(self):
...
+            Priorities.HIGH]
         for item, priority in zip(items, priorities):
             queue.push(priority, item)
 
-        items.reverse()
+        items = [4, 3, 1, 2]
```
ClayMav, line 25: items = items[::-1]

Utilities/priority_queue.py
```diff
@@ -12,6 +12,8 @@ class PriorityQueue():
         The priority queue.
     """
 
+    count = 987654
```
ClayMav: Can you justify this for me

ClayMav (changes requested):
Small stuff and some justification

Tests/priority_queue_test.py
Thomas-McKanna: That's the reason for the pull request. There are going to be situations where we may put a bunch of instructions on the queue that are of equal priority. We want instructions of equal priority to be read in the order they were put in the queue. The old implementation had it so that the most recent instruction (of equal priority) to be added to the queue was executed first.

Tests/priority_queue_test.py
Thomas-McKanna: Its not exactly a reversal of the list. Notice that 1 and 2 have been swapped. The list indicates the expected ordering of the items.

Utilities/priority_queue.py
Thomas-McKanna: I have realize that I did not choose this number correctly. A better one would be 100000. Here is the justification: this implementation of the priority queue works by appending the `priority.value` (and enum value) to the `count` variable, after which the `count` in incremented. In this way, instructions of equal priority are executed in the order that they were put into the queue.

For example, say we have two move instructions of `MEDIUM` priority. The first one gets a value of `'2' + '100000'` or `2100000` and the second one would get `'2' + '100001'` or `2100001`  associated with it. Since the queue is a min heap, the first movement would be read first, just like we want.

By choosing 100000 we could have up to 900000 instructions in the queue that will be executed in the right order. This is more than enough.

Tests/priority_queue_test.py
ClayMav: How does duplicating the low priority help with that

Tests/priority_queue_test.py
ClayMav: I see

Utilities/priority_queue.py
ClayMav: understood, you should definitely look into writing a quick priority heap down the line, it wouldn't be a difficult change to accomplish this.

Tests/priority_queue_test.py
Thomas-McKanna: In the test, the item with value 1 should be dequeued before the item with value 2, since they are equal priority (LOW) and the item with value 1 is enqueued first.

Utilities/priority_queue.py
```diff
@@ -12,6 +12,8 @@ class PriorityQueue():
         The priority queue.
     """
 
+    count = 900000
```
christopher-o-toole: don't hardcode configurable values, use constants.
```suggestion
    count = DEFAULT_COUNT
```

Utilities/priority_queue.py
```diff
@@ -27,7 +29,9 @@ def push(self, priority, item):
             The data being inserted into the queue.
 
         """
-        heapq.heappush(self._queue, (priority.value, item))
+        val = int('{}{}'.format(priority.value, PriorityQueue.count))
+        PriorityQueue.count = PriorityQueue.count + 1
```
christopher-o-toole: Use compound operators where applicable
```suggestion
        PriorityQueue.count += 1
```

Tests/priority_queue_test.py
christopher-o-toole: Since this is a test function, it's perfectly reasonable for two lows to be in the list here as it verifies that item priority is handled appropriately, and even if it were an erroneous case, then that's still something valid to test for.

christopher-o-toole (changes requested):
Pretty good, but there are a few minor changes to be made.

christopher-o-toole (approved):

ClayMav (approved):

PR #17, opened 2019-01-27
Adding the pixhawk params after a successful day of testing ✈️

25cent9:
I'm adding the flight ready params for the PR drone to the repo.


christopher-o-toole (approved):
Google Drive is preferred for this type of file.

ClayMav:
This is a binary file, for the time being, it doesn't make any sense to me to have it in here. We will definitely give this another discussion, but I feel like somewhere on the drive is where this should be. I see the point in having it in the repo, but I don't feel comfortable since there is no diff.

PR #19, opened 2019-02-06
Feature/reorganize

Thomas-McKanna:
This is the reorganize of the flight code according to Clay's diagram. During the process of reorganizing, I changed certain aspects of the flight code. Of note:

- The Instruction class has been rolled into the Task class (the old instructions class's sole purpose was to decode messages sent over the network into tasks, so the name Instruction was misleading)
- The Drone class now sub-classes dronekit.Vehicle (rather that containing a variable that was a dronekit.Vehicle). This simplified things. For example, the old code would have had to do `drone._vehicle.rangerfinder.distance` and new code can do `drone.rangefinder.distance`.
- The Drone class is now purely an interface to the drone. It does not calculate pid values or sleep the program. These sorts of thing were better suited for DroneController, which is where they now reside. This change will allow Tasks to be much more dynamic, but has required all Tasks to be rewritten.
- Added a task for takeoff, land, and hover (of note: movement not yet implemented, but this won't be hard)

This is a big one, but I'm hoping we can get done by the flight test this weekend.


flight/drone/drone.py
```diff
@@ -0,0 +1,186 @@
...
+        def listener(self, name, message):
+            """
+            The listener is called for messages that contain the string specified
+            in the decorator,passing the vehicle, message name, and the message.
+            """
+            self._optical_flow.time_usec = message.time_usec
+            self._optical_flow.sensor_id = message.sensor_id
+            self._optical_flow.flow_x = message.flow_x
+            self._optical_flow.flow_y = message.flow_y
+            self._optical_flow.flow_comp_m_x = message.quality
```
christopher-o-toole:
```suggestion
            self._optical_flow.flow_comp_m_x = message.flow_comp_m_x
```

flight/drone/drone.py
```diff
@@ -0,0 +1,186 @@
...
+
+    def __init__(self, *args):
+        super(Drone, self).__init__(*args)
+        self._id = None
+        self._logger = logging.getLogger(__name__)
+
+        self._optical_flow = OpticalFlow()
+
+        # Allow us to listen for optical flow dat
+        @self.on_message('OPTICAL_FLOW')
```
christopher-o-toole:
```suggestion
        @self.on_message(OPTICAL_FLOW_MESSAGE)
```

flight/drone/drone.py
```diff
@@ -0,0 +1,186 @@
...
+            self._optical_flow.flow_y = message.flow_y
+            self._optical_flow.flow_comp_m_x = message.quality
+            self._optical_flow.flow_comp_m_y = message.flow_comp_m_y
+            self._optical_flow.quality = message.quality
+            self._optical_flow.ground_distance = message.ground_distance
+
+            # Notify all observers of new message (with new value)
+            #   Note that argument `cache=False` by default so listeners
+            #   are updaed with every new message
+            self.notify_attribute_listeners('optical_flow', self._optical_flow)
```
christopher-o-toole:
```suggestion
            self.notify_attribute_listeners(OPTICAL_FLOW_MESSAGE.lower(), self._optical_flow)
```

flight/drone/drone.py
```diff
@@ -0,0 +1,186 @@
...
+    def id(self, identifier):
+        """Set the drone's id."""
+        self._id = identifier
+
+    def set_attitude(self, roll, pitch, yaw, thrust):
+        """Set the drones attitude.
+
+        Parameters
+        ----------
+        roll : double
```
christopher-o-toole: Technically Python has only a float type rather than double, but that's probably nitpicking.

flight/drone/drone.py
```diff
@@ -0,0 +1,186 @@
...
+        MAVLink_message (a DroneKit object)
+            Message which moves the drone at a certain velocity.
+
+        see http://python.dronekit.io/examples/guided-set-speed-yaw-demo.html#send-global-velocity
+        """
+        return self.message_factory.set_position_target_global_int_encode(
+            0,       # time_boot_ms (not used)
+            0, 0,    # target system, target component
+            mavutil.mavlink.MAV_FRAME_GLOBAL_RELATIVE_ALT_INT, # frame
+            0b0000111111000111, # type_mask (only speeds enabled)
```
christopher-o-toole: What's the reasoning for removing the Bitmask enum?

flight/drone/drone.py
```diff
@@ -0,0 +1,186 @@
...
+            The thrust value.
+        """
+        # Thrust >  0.5: Ascend
+        # Thrust == 0.5: Hold the altitude
+        # Thrust <  0.5: Descend
+        return self.message_factory.set_attitude_target_encode(
+            0, # time_boot_ms
+            1, # Target system
+            1, # Target component
+            0b00000000, # Type mask: bit 1 is LSB
```
christopher-o-toole: You don't really need to specify that 1 is the LSB here, also, there's no info as to what this bitmask means.

flight/drone/drone.py
```diff
@@ -0,0 +1,186 @@
...
+            self._to_quaternion(roll, pitch), # Quaternion
+            0, # Body roll rate in radian
```
christopher-o-toole:
```suggestion
            0, # Body roll rate in radians
```

flight/drone/drone.py
```diff
@@ -0,0 +1,186 @@
...
+            0, # Body pitch rate in radian
```
christopher-o-toole:
```suggestion
            0, # Body pitch rate in radians
```

flight/drone/drone.py
```diff
@@ -0,0 +1,186 @@
...
+            radians(yaw), # Body yaw rate in radian
```
christopher-o-toole:
```suggestion
            radians(yaw), # Body yaw rate in radians
```

flight/drone/optical_flow_attribute.py
```diff
@@ -0,0 +1,41 @@
...
+            flow_y=None, flow_comp_m_x=None, flow_comp_m_y=None, quality=None,
+            ground_distance=None):
+        """
+        OpticalFlow object constructor.
+        """
+        self.time_usec = time_usec
+        self.sensor_id = sensor_id
+        self.flow_x = flow_x
+        self.flow_y = flow_y
+        self.flow_comp_m_x = quality
```
christopher-o-toole:
```suggestion
        self.flow_comp_m_x = flow_comp_m_x
```

flight/drone/optical_flow_attribute.py
```diff
@@ -0,0 +1,41 @@
...
+        self.flow_comp_m_y = flow_comp_m_y
+        self.quality = quality
+        self.ground_distance = ground_distance
+
+    def __str__(self):
+        """
+        String representation used to print the OpticalFlow object.
+        """
+        return "OPTICAL_FLOW: time_usec={},sensor_id={},flow_x={},flow_y={}, \
```
christopher-o-toole: Consider making the format string a static class constant to reduce the longevity of this line.

flight/tasks/hover_task.py
```diff
@@ -0,0 +1,40 @@
...
+        self._count = duration * (1.0/c.DELAY_INTERVAL)
+
+    def perform(self):
+        # Get control value
+        zv = -self._pid_alt(self._drone.rangefinder.distance)
+        # Send 0 velocities to drone (excepting altitude correction)
+        self._drone.send_velocity(0, 0, zv)
+        self._count -= 1
+
+        if self._count <= 0:
```
christopher-o-toole: Unnecessary conditional, just do this:

```suggestion
        return self._count <= 0
```

flight/tasks/land_task.py
```diff
@@ -0,0 +1,27 @@
...
+            The drone being controlled.
+        """
+        super(LandTask, self).__init__(drone)
+        self._land_mode = VehicleMode(c.Modes.LAND.value)
+
+    def perform(self):
+        if not self._drone.mode == self._land_mode:
+            self._drone.mode = self._land_mode
+            return False
+        elif self._drone.armed:
```
christopher-o-toole: unnecessary conditional, just do
```suggestion
        return not self._drone.armed
```

test_scripts/test_hover.py
```diff
@@ -0,0 +1,10 @@
+from ..flight.drone.drone_controller import DroneController
+from ..flight import constants as c
+
+controller = DroneController(c.Drones.LEONARDO_SIM)
+
+controller.add_takeoff_task(c.DEFAULT_ALTITUDE)
+controller.add_hover_task(c.DEFAULT_ALTITUDE, 10, c.Priorities.MEDIUM)
```
christopher-o-toole: Don't hardcode constants, choose an approriate name for what `10` represents as a constant variable.

unit_tests/simple_imports.py
```diff
@@ -7,6 +7,6 @@ def import_parent():
 
 
 def import_distributor():
-    directory = parent_dir + r"\tools\data_distributor"
+    directory = os.path.join(parent_dir, 'tools', 'data_distributor')
```
christopher-o-toole, line 10: Make constants for the folder names here, as they may change in the next reorganization.

flight/tasks/task_base.py
```diff
@@ -0,0 +1,38 @@
...
+class TaskBase():
+    """A task the the drone can perform.
+
+    Responsible for implementing the core logic of the various actions that a
+    drone can take (ex. Movement, Follow, Heal, Decode). Must implement
+    perform().
+
+    Attributes
+    ----------
+    _drone: dronekit.vehicle
```
christopher-o-toole, line 13: Inconsistent documentation: you have not documented private attributes anywhere else. You probably should attempt to document all attributes in general, even if it's just an inline comment for a private attribute.

christopher-o-toole (changes requested):
Generally you should document each file with a docstring but I'll let it slide for now given our schedule. The documentation and use of constants was a bit inconsistent in some areas as well. Other than that, the majority of the code looked alright.

flight/drone/drone.py
Thomas-McKanna: good catch

flight/drone/drone.py
Thomas-McKanna: The string here has to exactly match the name of the mavlink attribute. See https://mavlink.io/en/messages/common.html#OPTICAL_FLOW. on_message() is a dronekit.Vehicle method.

flight/drone/drone.py
Thomas-McKanna: Right. I tend to forget that. I will change it.

Thomas-McKanna:
Chris, I will add docstrings for each file to the Trello todo list.

flight/drone/drone.py
Thomas-McKanna: They got lost in the shuffle. I have added a MavBitmasks class in the constant folder and made a docstring that explains the meaning of the bitmasks and provides links to the fields the bitmask controls.

unit_tests/simple_imports.py
Thomas-McKanna: I think I will leave this one up to Cole, because it looks like he is finding the parent directory using some os function. I wouldn't know what constant to make it.

flight/tasks/task_base.py
Thomas-McKanna: I have added documentation for all variables.

ClayMav (changes requested):
* `test_scripts` should not exist, and its contents should be in `unit_tests`.
* There should be no need for an `__init__.py` in the root directory.
* There are no configuration parameters in the flightconfig.py, I would figure at least a few things should be configurable from flight to flight.
I don't think theres anything I really want to nitpick in the code here.

flight/drone/drone.py
```diff
@@ -0,0 +1,195 @@
...
+    def __init__(self, *args):
+        super(Drone, self).__init__(*args)
+        self._id = None
+        self._logger = logging.getLogger(__name__)
+
+        self._optical_flow = OpticalFlow()
+
+        # Allow us to listen for optical flow dat
+        @self.on_message(c.OPTICAL_FLOW_MESSAGE)
+        def listener(self, name, message):
```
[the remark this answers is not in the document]
Thomas-McKanna: We may use this it some point. There for future convenience.

christopher-o-toole (dismissed):

ClayMav (approved):

christopher-o-toole (approved):

PR #20, opened 2019-02-08
Feature/exit task

Thomas-McKanna:
This pull request solves two problems:

1. Previous implementation used LandTask as a flag to shut down the controller. This was an issue because it meant we couldn't land and then take off again in the same session (which would require lots of restarting the soon-to-come server during flight tests). To solve this problem, an ExitTask class was added which now serves as a flag to shut down the controller.
2. A bug existed where if a task was interrupted by a task of higher priority, it would be moved back in the queue, and once it was at the front again, it restarted itself from the beginning (ex. moving right for 7 out of 10 seconds and then interrupted would result in the drone eventually trying to flight right for an additional 10 seconds). New implementation discards interrupted tasks (may not be the final solution, but it eliminates the safety concerns of the old implementation).

In addition, I have added a simple command line interface for dynamically controlling the drone (located in AI folder).


flight/AIs/test_linear_move.py
```diff
@@ -10,7 +10,14 @@
 controller.add_takeoff_task(f.DEFAULT_ALTITUDE)
 controller.add_hover_task(
     f.DEFAULT_ALTITUDE, HOVER_DURATION, c.Priorities.MEDIUM)
+"""
```
ClayMav: This file doesn't run a linear movement

flight/AIs/user_command.py
```diff
@@ -0,0 +1,73 @@
...
+from ... import flightconfig as f
+
+def input_loop(controller):
+    """Simple command line interface to drone controller.
+
+    exit: lands and terminates program
+    land: land <priority>
+    hover: hover <duration> <priority>
+    takeoff: currently not used for safety reasons
+    move: move <FOWARD,BACK,LEFT,RIGHT> <duration> <priority>
```
ClayMav:
```suggestion
    move: move <FORWARD,BACK,LEFT,RIGHT> <duration> <priority>
```

ClayMav: also you have `BACK` here and `BACKWARD` below

flight/AIs/user_command.py
```diff
@@ -0,0 +1,73 @@
...
+
+    Notes
+    -----
+    Priority is one of "high", "med", or "low".
+    """
+    while True:
+        command = raw_input('> ')
```
ClayMav:
```suggestion
        command = raw_input('> ').lower()
```

flight/AIs/user_command.py
```diff
@@ -0,0 +1,73 @@
...
+        command = command.lower()
```
ClayMav:
```suggestion
```

flight/AIs/user_command.py
```diff
@@ -0,0 +1,73 @@
...
+            controller.add_land_task(priority)
+        elif len(commands) == 3 and commands[0] == "hover":
+            priority = get_priority(commands[2])
+            controller.add_hover_task(
+                f.DEFAULT_ALTITUDE, float(commands[1]), priority)
+        elif len(commands) == 2 and commands[0] == "takeoff":
+            controller.add_takeoff_task(float(commands[1]))
+        elif len(commands) == 4 and commands[0] == "move":
+            priority = get_priority(commands[3])
+            if commands[1] == "forward":
```
ClayMav: this whole if else chain should be a function

flight/AIs/user_command.py
```diff
@@ -0,0 +1,73 @@
...
+                dir = c.Directions.FORWARD
```
ClayMav: dir is reserved, you are [clobbering](https://en.wikipedia.org/wiki/Clobbering) it. change the variable name to `direction`

flight/drone/drone.py
```diff
@@ -113,6 +114,33 @@ def send_velocity(self, north, east, down):
...
+        Only guided mode is currently supported.
+        """
+        self.mode = VehicleMode(mode)
+
+        self._logger.info('Arming...')
+        while not self.armed:
+            self.armed = True
+            time.sleep(c.ARM_RETRY_DELAY)
+
+        status_msg = 'Failed to arm' if not self.armed else 'Armed'
```
ClayMav: Should be
```python
if self.armed:
    self._logger.info("Armed")
else:
    self._logger.error("Failed to arm")
```
What you have now is not very readable

ClayMav (changes requested):

flight/AIs/test_linear_move.py
Thomas-McKanna: I had accidentally left some code commented out. I have un-commented and cleaned it up.

flight/AIs/user_command.py
Thomas-McKanna: Thanks. I have also added the command formatting for takeoff, since it is now usable.

flight/AIs/user_command.py
Thomas-McKanna: I have added a get_direction() function.

flight/AIs/user_command.py
Thomas-McKanna: I have learned something new. The change has been made.

flight/AIs/user_command.py
```diff
@@ -0,0 +1,82 @@
...
+    if string == "high":
+        priority = c.Priorities.HIGH
+    elif string == "med":
+        priority = c.Priorities.MEDIUM
+    elif string == "low":
+        priority = c.Priorities.LOW
+
+    return priority
+
+def get_direction(string):
```
christopher-o-toole: This is inefficient since it has O(n) complexity in the worst case where n is the number of conditional statements. Also this does not do any input validation. In fact, I don't think it's necessary to support string directions at all, and even if it was, you could just do

```Python
converted_form = None
key = direction.upper()
if hasattr(c.Directions, key):
     converted_form = getattr(c.Directions, key)
else:
     raise TypeError('Expected value for type {}, got {}.'.format(type(c.Directions),  direction))
```

flight/AIs/user_command.py
```diff
@@ -0,0 +1,82 @@
...
+                continue
+            else:
+                controller.add_linear_movement_task(
+                    direction, float(commands[2]), priority)
+        else:
+            print "> Unknown command"
+
+def get_priority(string):
+    """Gets priority level from a string."""
+    priority = None
```
christopher-o-toole: Same problem as mentioned further in this review, either use getattr() or have a map which defines a function from a string to a priority enum value.

flight/AIs/user_command.py
```diff
@@ -0,0 +1,82 @@
+import threading
+
+from ..drone.drone_controller import DroneController
+from .. import constants as c
+from ... import flightconfig as f
```
christopher-o-toole, line 7: This works? Seems strange to me but maybe I don't know relative import syntax that well.

christopher-o-toole: Note the ```...``` is what I'm referring to.

flight/AIs/user_command.py
```diff
@@ -0,0 +1,82 @@
...
+    land: land <priority>
+    hover: hover <duration> <priority>
+    takeoff: takeoff <altitude>
+    move: move <FOWARD,BACKWARD,LEFT,RIGHT> <duration> <priority>
+
+    Notes
+    -----
+    Priority is one of "high", "med", or "low".
+    """
+    while True:
```
christopher-o-toole, line 134: When does this terminate?

flight/AIs/user_command.py
```diff
@@ -0,0 +1,82 @@
...
+        command = raw_input('> ').lower()
```
christopher-o-toole: Also
```suggestion
        command = raw_input(PROMPT_FOR_COMMAND).lower()
```
Use constants for configurable values.

flight/AIs/user_command.py
```diff
@@ -0,0 +1,82 @@
...
+            controller.add_hover_task(
+                f.DEFAULT_ALTITUDE, float(commands[1]), priority)
+        elif len(commands) == 2 and commands[0] == "takeoff":
+            controller.add_takeoff_task(float(commands[1]))
+        elif len(commands) == 4 and commands[0] == "move":
+            priority = get_priority(commands[3])
+            direction = get_direction(commands[1])
+
+            if priority is None or direction is None:
+                print "> Invalid movement command"
```
christopher-o-toole:
```suggestion
                print '{}Invalid movement command".format(PROMPT_FOR_COMMAND)
```

flight/AIs/user_command.py
```diff
@@ -0,0 +1,82 @@
...
+                continue
+            else:
+                controller.add_linear_movement_task(
+                    direction, float(commands[2]), priority)
+        else:
+            print "> Unknown command"
```
christopher-o-toole:
```suggestion
            print "{}Unknown command".format(PROMPT_FOR_COMMAND)
```

flight/AIs/user_command.py
```diff
@@ -0,0 +1,82 @@
...
+    takeoff: takeoff <altitude>
+    move: move <FOWARD,BACKWARD,LEFT,RIGHT> <duration> <priority>
+
+    Notes
+    -----
+    Priority is one of "high", "med", or "low".
+    """
+    while True:
+        command = raw_input('> ').lower()
+        commands = command.split()
```
christopher-o-toole: See discord DM's for further comments, however you should also probably use an enum for the command values.

flight/AIs/user_command.py
```diff
@@ -0,0 +1,82 @@
...
+    elif string == "right":
+        direction = c.Directions.RIGHT
+
+    return direction
+
+# Make the controller object
+controller = DroneController(c.Drones.LEONARDO_SIM)
+
+# Make a thread whose target is a command line interface
+input_thread = threading.Thread(target=input_loop, args=(controller,))
```
christopher-o-toole: non-daemon thread may prevent app exit (due to infinite while loop)

christopher-o-toole (changes requested):
The skeleton is there, but there's some definite cleanup to do. See requested changes for details.

flight/AIs/user_command.py
ClayMav: That's a fallacy of Big-O notation, this is a constant time operation. However, that is a novel solution.

flight/AIs/user_command.py
ClayMav: It shouldn't unless the user wants it to end, so either by ctrl-c or exit

flight/AIs/user_command.py
Thomas-McKanna: Its weird but it works - as long as you run the program as a module. When Clay and I were debugging his networked command program (which wasn't run as a module), we had to change the relative imports for it to work.

flight/AIs/user_command.py
Thomas-McKanna: Yeah, what Clay wrote are the two conditions that will cause it to terminate. Is there a better way to go about this?

christopher-o-toole (approved):

flight/AIs/user_command.py
ClayMav: We will in the future turn everything into module based imports instead of relative, relative is almost a capital crime

ClayMav (approved):

PR #21, opened 2019-02-09
Added requirements.txt file.

Thomas-McKanna:
I _think_ this is everything that is needed. I know that some of the packages used for tools are not in here, and I can't find a requirements.txt in the tools folder. Cole should be able to add whatever else is needed.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,7 @@
+dronekit == 2.9.1
+pymavlink == 2.2.1
+mavproxy == 1.7.7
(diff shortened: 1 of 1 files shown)


ClayMav (approved):
Way too much code in this PR, next time, a little less in the PR please

christopher-o-toole (approved):
Approving on the condition that you ensure this file works as expected.

25cent9:
Can you write it in `LaTex`??

PR #22, opened 2019-02-09
Network Flight support

ClayMav:
Includes the new `flight.py`, a web server which allows for controlling the drone via HTTP.
Also included is an Ansible playbook which will grab specified commit/branch, and pull it onto the drone, and then run the flight code. All of this needs to be tested and cleaned up.


ClayMav:
Admittedly, some files were changed that didn't need updates, but they better comply with pep8 now, so no harm no foul

flight/drone/drone_controller.py
```diff
@@ -16,7 +19,7 @@
 from ..tasks.takeoff_task import TakeoffTask
 from ..utils.priority_queue import PriorityQueue
 from ..utils.timer import Timer
-from ... import flightconfig as f
+
```
Thomas-McKanna: drone_controller.py needs to use the flight config constants for things like safety checks. I don't really understand these relative imports and why flightconfig is the only one that is causing issues.
```suggestion
import flightconfig as fc
```

flight.py
```diff
@@ -0,0 +1,170 @@
...
+
+
+def debug_add_task(controller, command, meta):
+    if "priority" in meta:
+        priority = get_priority(meta["priority"])
+    if "direction" in meta:
+        direction = get_direction(meta["direction"])
+
+    if command == "exit":
+        controller.add_land_task(c.Priorities.HIGH)
```
Thomas-McKanna: This would be a good place to put the new ExitTask, which causes the drone controller to shut down. 
```suggestion
        controller.add_exit_task(c.Priorities.HIGH)
```

flight.py
```diff
@@ -0,0 +1,170 @@
...
+            'Expected value for type {}, got {}.'.format(
+                type(c.Directions), direction))
+
+
+def flask_thread(args, controller):
+    blueprints = create_routes(args, controller)
+    app = Flask(__name__)
+    for blueprint in blueprints:
+        app.register_blueprint(blueprint)
+    app.run("127.0.0.1", 8000)
```
Thomas-McKanna: I noticed on the Flask website that Flask.run() has "threaded mode" enabled by default (http://flask.pocoo.org/docs/1.0/api/?highlight=run#flask.Flask.run). Would this mean that calling app.run() would spawn a new thread for itself to run on? If that's the case, we may not need to create a flask_thread at all.

flight.py
```diff
@@ -0,0 +1,170 @@
...
+
+
+def main():
+    # Get Command Line Arguments
+    args = parse_args()
+
+    # Establish controller
+    controller = DroneController(c.Drones.LEONARDO_SIM)
+    # Run Flask server on seperate thread
+    server_thread = threading.Thread(
```
Thomas-McKanna: If the drone controller ever shuts down, I don't see a way for the Flask thread to shutdown, which would cause the program to hang. I know you have `sys.exit(0)` after `controller.run()`, but it won't cause the flask thread to shut down. You will need to make the `server_thread` daemon if you want that behavior.

flight.py
```diff
@@ -0,0 +1,170 @@
...
+        target=flask_thread, args=(args, controller))
+
```
Thomas-McKanna:
```suggestion
        server_thread.daemon = True
```
This will cause the server to abruptly shutdown. If that isn't acceptable, we will have to find an alternative way (probably using threading.Events).

flight_clients/cli/client.py
```diff
@@ -0,0 +1,40 @@
...
+URL = 'http://{}:8000/commands/'.format(sys.argv[1])
+
+
+def gen_req(command, **kwargs):
+    resp = requests.post(URL, data={"command": command, "meta": kwargs})
+    if resp.status_code != requests.codes.ok:
+        logging.error(resp.text)
+
+
+while True:
```
Thomas-McKanna: If you think it would be good, you could use the Command class which is used in user_command.py. The benefit is that it does type checking for each of the parameters of each command, and it prevents an ever growing list of elif statements.

Thomas-McKanna (changes requested):
Nice work! We will have to work out the bugs either in person or in a discord call, because when I try to run it now, I'm getting some error about NoneType not being iterable (and there may be more bugs beyond it). I'm sure it won't be hard to get past those small bugs.

It seems like you are doing command validation on the server rather than the client. For instance, if the user gave a faulty direction, it is caught on the server with an `InvalidDirectionException`. It would be good to move the burden of input validation from the server to the client. This could be done by using the `Command` base class used into `user_command.py` in `client.py`.

flight.py
ClayMav: After looking into it, I am not positive that threaded mode runs the flask server on another thread, but handles each new connection on a separate thread. I could be wrong, but we will want to test that.

flight/drone/drone_controller.py
ClayMav: My mistake, I don't believe this was an intentional change.

flight_clients/cli/client.py
ClayMav: I do agree that in terms of style, `if elif elif elif` is not ideal, and I will think about a good way to solve that, but at the current moment, I am not convinced that your solution is any more elegant, it seems much less readable. For @christopher-o-toole's sake however, using the dictionary to solve this makes it instant lookup(O(1)), which is definitely... something... (maybe nitpicking)

Thomas-McKanna: I agree that the way I did it is barely readable. The thing I dislike the most about the elif elif elif way of doing it is that there is lots of indexing by constants into the args variable, and because each command can have a different ordering of arguments, it makes things hard to understand. Maybe if we standardized the way a command is sent, it would help.
```python
COMMAND_NAME = 0
COMMAND_PRIORITY = 1
CUSTOM_ARGS_START = 2
...
name = args[COMMAND_NAME]
priority = args[COMMAND_PRIORITY]
custom_args = args[CUSTOM_ARGS_START:]
if name == "move":
    process_movement_args(custom_args)
    controller.add_movement_task(*custom_args, priority)
... 

```

flight.py
```diff
@@ -0,0 +1,225 @@
...
+    args = parser.parse_args()
+
+    if not args.debug and not args.routine:
+        logging.error("Please specify routine.")
+        sys.exit(1)
+    return args
+
+
+def create_routes(args, controller):
```
Thomas-McKanna: Will this be used in the future?

flight.py
```diff
@@ -0,0 +1,225 @@
...
+def get_direction(direction):
+    return get_enum(c.Directions, direction)
+
+
+def get_enum(enum, key):
+    item = key.upper()
+    if hasattr(enum, item):
+        return getattr(enum, item)
+    else:
+        raise InvalidDirectionException(
```
Thomas-McKanna: The exception is specific to direction, even though it's being used for priority too.

flight.py
```diff
@@ -0,0 +1,225 @@
...
+            'Expected value for type {}, got {}.'.format(type(enum), item))
+
+
+def start_ai(controller, routine):
+    importlib.import_module("flight.AIs.{}".format(routine))
```
Thomas-McKanna: Is the `controller` parameter being used here?

flight.py
```diff
@@ -0,0 +1,225 @@
...
+            return debug_add_task(controller, command, meta)
+        except (InvalidDirectionException, InvalidPriorityException) as err:
+            return err
+        except Exception as err:
+            logging.error("Unexpected error, killing drone" + str(err))
+            debug_add_task(controller, "exit", {})
+            return False
+    elif args.routine:
+        # If in production mode, starting or stopping drone is only option
+        # get push data and check if start or kill
```
Thomas-McKanna: Correct me if I'm wrong here, but is this comment supposed to say "get post data" instead of "get push data"?

flight.py
```diff
@@ -0,0 +1,225 @@
...
+        command = data["command"].lower()
+        if command == "start":
+            # start routine
+            start_ai(controller, args.routine)
+        elif command == "kill":
```
Thomas-McKanna: What if command is not equal to "start" or "kill"?

flight.py
```diff
@@ -0,0 +1,225 @@
...
+            # force into land
+            kill_ai(controller)
+
+        return "routine " + args.routine + " in progress"
+
+
+def tcp_thread(args, controller):
+    sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    # set_keepalive_osx(s)
```
Thomas-McKanna: Are the keepalive functions for the client? If so, would it be hard to add windows as an option? The laptop I always bring during testing sessions runs windows. Also, could the OS be a command line arg?

flight/drone/drone.py
```diff
@@ -114,7 +114,7 @@ def send_velocity(self, north, east, down):
         msg = self._make_velocity_message(north, east, down)
         self.send_mavlink(msg)
 
-    def arm(self, mode=c.Modes.GUIDED.value):
+    def arm(self, mode=c.Modes.GUIDED):
```
Thomas-McKanna, line 119: I'm anticipating all the removals of `.value` is going cause the code to not work on my machine. Your going to have to tell me (and perhaps add to the README) how I should start the flight code. Something to do with pipenv, right?

tools/flight_cli/client.py
```diff
@@ -0,0 +1,80 @@
...
+        params = input("> ").split()
+        try:
+            command = params[0]
+            logging.info(command)
+            if command == "exit":
+                gen_req(s, command)
+                s.shutdown(socket.SHUT_WR)
+                s.close()
+                return
+            elif command == "land":
```
Thomas-McKanna: Can you put a comment or docstring somewhere that lists out the available commands and the order of parameters expected for each command?

Thomas-McKanna (commented):
I added some comments. Most are questions or small things. I'm thinking this is real close to being merged.

flight.py
ClayMav: For the drone, so there won't be any OS specific stuff.

flight/drone/drone.py
ClayMav: Will add to `README.md`, the commands are as follows:
```bash
pip install pipenv --user
pipenv install --skip-lock
pipenv shell
```
then run as you will

ClayMav:
Currently if the client tells the drone to die, the drone dies, but the socket doesn't close on the server. Needs to be fixed.

PR #24, opened 2019-02-13
Servo Controller

jozgd:


ClayMav:
So this PR has a huge problem, when you started work on this you checked out the feature/reorganize branch and didn't create a new branch. What that caused is that you are now making laser changes to the feature/reorganize branch. You are also not in a good situation with regards to the "state tree" of the repository.

What I suggest you do, is:
1. Take your laser changes, copy the new files used for them
2. checkout to develop and pull all new changes
3. create a new branch with a correct name
4. switch to that new branch
5. place all the new files in to this new branch
6. test your changes
7. commit and push
8. Open a new PR with the correct branch.


In the future, make sure you are always branching from develop, and never from a feature branch unless you need something it has. It is very important to work on your own branch so you don't get lost in the "state tree". This PR is going to close now, but once you get this stuff fixed, I have no problem reviewing it again!

PR #26, opened 2019-02-14
Updated to check if the drone is armable before arming

jozgd:
The previous code was potentially trying to arm the drone before the drone was able to be armed and this was causing a safety error.


ClayMav (approved):
clean af

christopher-o-toole (approved):

PR #27, opened 2019-02-14
Incorporated logging/rtg into existing flight code.

Thomas-McKanna:
This pull request adds capabilities of logging and real-time graphing (RTG) into the flight code. The logger will receive all drone attributes, while the RTG will select a subset of available attributes to graph. Note: this pull request is primarily to allow us to begin logging flights, not using the grapher (which was not yet ready to test before this pull request). 

Added features:

- DroneController has gained a DataSplitter object, which is passed to a Timer to send data at regular intervals.
- DroneController has a new function, `_gather_data()`, which is where sensor data is actually queried from the drone. Let me know if the use of `eval` is distasteful here.
- A dictionary called ATTRIBUTE_TO_FUNCTION has been added to the constants folder, which maps a predefined set of attribute names to the function that gets that attribute off of the drone.

The code in the request was tested and produced good logs of data. No negative effect to flight performance was observed. Note: running the grapher for just 30 seconds generated 30MB of data. This could be a problem, since logs will (in the current implementation) be stored on the drone itself.

### DIFF
--- a/flight/drone/drone_controller.py
+++ b/flight/drone/drone_controller.py
@@ -276,13 +276,30 @@ class DroneController(object):
         Used to send a dictionary of data to a DataSplitter object, which logs
         and potential graphs the data.
 
+        Some of the attributes we graph require pulling class attributes:
+            Ex. rangefinder.distance
+        While for other just the class itself is fine:
+            Ex. airspeed
+        In addition, some attributes require you to index into them for data:
+            Ex. velocity[0] => x velocity
+
+        Hence the ugly if elif elif structure beow.
+
         Returns
         -------
         dict
         """
         data = {}
-        for attribute, function_suffix in c.ATTRIBUTE_TO_FUNCTION.iteritems():
-            data[attribute] = eval("self._drone.{}".format(function_suffix))
+        for attr_name, attr in c.ATTRIBUTE_TO_FUNCTION.iteritems():
+            if len(attr) == 1:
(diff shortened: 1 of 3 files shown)


christopher-o-toole (dismissed):
Looks good to me.

flight/drone/drone_controller.py
```diff
@@ -243,3 +267,22 @@ def _land(self):
...
+        Used to send a dictionary of data to a DataSplitter object, which logs
+        and potential graphs the data.
+
+        Returns
+        -------
+        dict
+        """
+        data = {}
+        for attribute, function_suffix in c.ATTRIBUTE_TO_FUNCTION.iteritems():
+            data[attribute] = eval("self._drone.{}".format(function_suffix))
```
ClayMav: I don't think this is gonna fly. Is there no other way to solve this problem without using `eval`?

ClayMav (changes requested):

flight/drone/drone_controller.py
Thomas-McKanna: I can't think of another way to do it that won't require the constants folder somehow having a reference to the drone. Alternatively, the dictionary ATTRIBUTE_TO_FUNCTION could be a constant member variable of DroneController, which would avoid the eval situation, but it would add some messiness to DroneController.py. Thoughts?

ClayMav (approved):

christopher-o-toole (approved):

PR #29, opened 2019-02-18
Added acceptable altitude range to move and hover.

Thomas-McKanna:
Added a constant for ACCEPTABLE_ALTITUDE_DEVIATION (set at 0.1 for now) which allows movement tasks and hover tasks to only correct movement if altitude has drifted beyond the acceptable range (in this case 0.1 meters from the target altitude). 

I also added the capability of setting and altitude for movement tasks, since previously it always went to the default altitude, which might not be want we want 100% of the time.

Also added some documentation to LinearMovementTask that was missing from before.

Also adjusted the PID values so that corrections that do happen are not so severe (I think this was contributing the the "hopping" behavior we saw in real-life tests).


ClayMav (approved):

flight/tasks/hover_task.py
```diff
@@ -3,7 +3,7 @@
 from task_base import TaskBase
 from .. import constants as c
 
-KP = 1
```
christopher-o-toole, line 6: Again, these constants should be documented.

christopher-o-toole (changes requested):
I think this is a good improvement relative to our previous method of control, however you should document  the PID gain value constants to make it more readable to those less familiar with the flight code.

ClayMav (changes requested):
Just what Chris said

flight/tasks/hover_task.py
Thomas-McKanna: I fixed this issue in the `feature/imports` pull request.

christopher-o-toole (approved):

ClayMav (approved):

ClayMav:
Lets get this merged @Thomas-McKanna

PR #30, opened 2019-02-19
Feature/imports

Thomas-McKanna:
Curious to see what you all think about this pull request. First and foremost, this pull request removes all relative imports.

Second, how tasks are imported has changed. Every new task added was resulting in another `from tasks.x.y.z import Z` being added to `dronecontroller.py`. I moved the contents of the `task` folder into its `__init__.py`, which greatly simplified how tasks can be imported. It is similar to how dronekit does it (https://github.com/dronekit/dronekit-python/blob/master/dronekit/__init__.py).


flight/drone/drone.py
```diff
@@ -133,8 +133,7 @@ def arm(self, mode=c.Modes.GUIDED.value):
 
         self._logger.info('Arming...')
         while not self.armed:
-            if self.is_armable:
-                self.armed = True
+            self.armed = True
```
christopher-o-toole, line 141: Is there a particular reason why this is preferred over checking if the drone is armable?

flight/tasks/__init__.py
```diff
@@ -0,0 +1,207 @@
+import abc
```
christopher-o-toole: docstring should be added to all brand new files from here on out

flight/tasks/__init__.py
```diff
@@ -0,0 +1,207 @@
...
+
+from simple_pid import PID
+from dronekit import VehicleMode
+
+from flight import constants as c
+from flight import flightconfig as f
+from flight.drone.exceptions import EmergencyLandException
+
+KP = 1
```
christopher-o-toole: we need a comment for each of the constants in this group indicating what they are/what their purpose is.

flight/tasks/__init__.py
```diff
@@ -0,0 +1,207 @@
...
+        """Initialize a task for terminating control.
+
+        Parameters
+        ----------
+        drone : dronekit.Vehicle
+            The drone being controlled.
+        """
+        super(Exit, self).__init__(drone)
+
+    def perform(self):
```
christopher-o-toole: Missing function docstring.

flight/tasks/__init__.py
```diff
@@ -0,0 +1,207 @@
...
+            Target altitude to maintain during hover.
+        duration : float
+            How many seconds to hover for.
+        """
+        super(Hover, self).__init__(drone)
+        self._duration = duration
+        self._pid_alt = PID(KP, KI, KP, setpoint=altitude)
+        self._count = duration * (1.0/c.DELAY_INTERVAL)
+
+    def perform(self):
```
christopher-o-toole: Missing function docstring

flight/tasks/__init__.py
```diff
@@ -0,0 +1,207 @@
...
+
+        Parameters
+        ----------
+        drone : dronekit.Vehicle
+            The drone being controlled.
+        """
+        super(Land, self).__init__(drone)
+        self._land_mode = VehicleMode(c.Modes.LAND.value)
+
+    def perform(self):
```
christopher-o-toole: Missing function docstring.

flight/tasks/__init__.py
```diff
@@ -0,0 +1,207 @@
...
+        self._pid_alt = PID(KP, KI, KP, setpoint=f.DEFAULT_ALTITUDE)
+        self._count = duration * (1.0/c.DELAY_INTERVAL)
+        velocities = []
+        for v in direction.value:
+            velocities.append(v * f.DEFAULT_SPEED)
+        self._vx = velocities[0]
+        self._vy = velocities[1]
+        self._vz = velocities[2]
+
+    def perform(self):
```
christopher-o-toole: Missing function docstring.

flight/tasks/__init__.py
```diff
@@ -0,0 +1,207 @@
...
+        Parameters
+        ----------
+        drone : dronekit.Vehicle
+            The drone being controlled
+        altitude : float
+            How many meters off the ground to take off to.
+        """
+        super(Takeoff, self).__init__(drone)
+        self._target_alt = altitude
+    def perform(self):
```
christopher-o-toole: Missing function docstring.

christopher-o-toole (changes requested):
I do think that getting rid of relative imports is ideal maintainability wise, however it may be somewhat more inconvenient to maintain many tasks which are all thrown into the same file. Perhaps consider having a tasks module folder in which that module's `__init__.py` imports all officially released tasks into the relevant namespace.

flight/drone/drone.py
Thomas-McKanna: Yeah, so it turns out that drone.vehicle.is_armable is just a bugged value (see https://github.com/dronekit/dronekit-python/issues/695 and https://discuss.ardupilot.org/t/is-armable-parameter-doesnt-change-dronekit-python/21574), or its just not being used right. It always returns False, and so it prevents takeoff.

I had John submit a pull request to add checking for is_armable, but we never actually tested it (not good).

flight/drone/drone_controller.py
```diff
@@ -8,16 +8,14 @@
...
-from ..tasks.land_task import LandTask
-from ..tasks.linear_movement_task import LinearMovementTask
-from ..tasks.takeoff_task import TakeoffTask
-from ...tools.data_distributor.data_splitter import DataSplitter
-from ..utils.priority_queue import PriorityQueue
-from ..utils.timer import Timer
-from ... import flightconfig as f
+
+from flight import constants as c
+import flight.tasks
```
ClayMav:
```suggestion
from flight.tasks import Hover, Takeoff, LinearMovement, Land, Exit
```

user_command.py
```diff
@@ -55,7 +55,7 @@ def matches(self, *args):
         """
         if len(args) != len(self._expected_order):
             raise TypeError('Expected {} arguments, got {}.'.format(
-                len(args),  len(self._expected_order)))
+                len(args),  len(self._expected_order)))#might need to flip these
```
ClayMav: Yeah this should be flipped fam

ClayMav (changes requested):
Hit a fix fam, then we good dawg

Thomas-McKanna:
I did the requested changes. In addition I have done the following:

- Added docstrings to _all_ files which I have written the majority of
- Fixed bug with giving a hover command in user_command.py
- Renamed user_command.py to test_cli.py

Code changes have been tested and no problems were encountered.

christopher-o-toole (dismissed):
Much better.

ClayMav (changes requested):
I am happy but Codacy is not, just fix some of those Codacy complaints and I will approve

ClayMav (dismissed):

christopher-o-toole (dismissed):

christopher-o-toole (approved):

ClayMav (approved):

PR #32, opened 2019-02-27
feature/Yaw

jozgd:
Added the ability to yaw the drone from a Yaw task.


flight/drone/drone_controller.py
```diff
@@ -212,6 +212,24 @@ def add_exit_task(self, priority=c.Priorities.HIGH):
         """
         new_task = Exit(self._drone)
         self._task_queue.push(priority, new_task)
+    
+    def add_yaw_task(self, heading, priority=c.Priorities.MEDIUM, yaw_speed=0, yaw_direction=1, relative=True):
```
ClayMav:
```suggestion
    def add_yaw_task(self, heading, priority=c.Priorities.MEDIUM):
```

flight/tasks/yaw_task.py
```diff
@@ -0,0 +1,57 @@
...
+    _new_heading : int
+        Stores the given heading after modding it by 360 to keep it in range.
+    _yaw_speed : int
+        The speed in degrees/sec that the drone performs the yaw.
+    _yaw_direction : int
+        The direction which you force the drone rotoate: -1 for counterclockwise, and 1 for clockwise.
+    __relative : bool
+        Stores whether the given heading is relative or absolute: True means relative and False means absolute.
+    """
+    def __init__(self, drone, heading, yaw_speed=0, yaw_direction=1, relative=True):
```
ClayMav:
```suggestion
    def __init__(self, drone, heading):
```

flight/tasks/yaw_task.py
```diff
@@ -0,0 +1,57 @@
...
+            The degrees/sec that the drone performs the yaw: defaults to 0.
+        yaw_direction : int
+            The direction which the drone should yaw, same as _yaw_direction: defaults to 1.
+        relative : bool
+            The status whether the passed heading is relative or absolute, same as _relative: defaults to True.
+        """
+        super(Yaw, self).__init__(drone)
+        self._has_started = False
+        self._new_heading = heading%360
+        self._yaw_speed = yaw_speed
```
ClayMav: Get yaw speed with a constant in config.py

flight/tasks/yaw_task.py
```diff
@@ -0,0 +1,57 @@
...
+        self._yaw_direction = yaw_direction
```
ClayMav: I don't think there should be a direction since relative doesn't exist now

flight/tasks/yaw_task.py
```diff
@@ -0,0 +1,57 @@
...
+        self._relative = relative
```
ClayMav: I don't think that relative is necessary now, we will relativize our code later

flight/tasks/yaw_task.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def perform(self):
+        """Do one iteration of logic for yawing the drone."""
+        if not self._has_started:
+            self.start_heading = self._drone.heading
+            self._drone.send_yaw(self._new_heading, self._yaw_speed, self._yaw_direction, self._relative)
+            self._has_started = True
+        elif ((not self._relative and abs(self._drone.heading - self._new_heading) > DEGREE_BUFFER) or
```
ClayMav: Very hard to read my mans, might want to make this more readable by turning into variables

ClayMav (approved):

ClayMav (changes requested):
I approved when i meant to deny

flight/tasks/yaw_task.py
```diff
@@ -0,0 +1,52 @@
...
+        self._yaw_direction = 1 #defaulted to 1 (clockwise)
+        self._relative = True #defaulted to always be relative
+
+    def perform(self):
+        """Do one iteration of logic for yawing the drone."""
+        if not self._has_started:
+            self.start_heading = self._drone.heading
+            self._drone.send_yaw(self._new_heading, self._yaw_speed, self._yaw_direction, self._relative)
+            self._has_started = True
+        self._is_not_finished = (abs(self._drone.heading - self.start_heading) < self._new_heading - DEGREE_BUFFER)
```
ClayMav: This is a syntax error. Make this go at the top of this function, and it doesn't need to be `self._is_not_finished`, just make it `is_not_finished` since it only gets referenced in this function

ClayMav (changes requested):
Codacy has some complaints, and I have listed my main one

flight/tasks/yaw_task.py
Thomas-McKanna: @ClayMav This issue is that this function gets called many times. Specifically, it gets called every time that this task is the current task and update() is called from the drone controller loop. This is to prevent a Task from blocking the control loop. Tasks need some what to keep track of their state across multiple calls to perform and using bools is one of the methods we have used for that.

flight/drone/drone.py
```diff
@@ -119,6 +119,12 @@ def send_velocity(self, north, east, down):
         msg = self._make_velocity_message(north, east, down)
         self.send_mavlink(msg)
 
+    def send_yaw(self, heading, yaw_speed=0, yaw_direction=1, relative=False):
```
christopher-o-toole:
```suggestion
    def send_yaw(self, heading, yaw_speed=DEFAULT_YAW_SPEED, yaw_direction=YawDirection.CLOCKWISE, relative=False):
```

flight/drone/drone.py
```diff
@@ -205,3 +211,16 @@ def _make_attitude_message(self, roll, pitch, yaw, thrust):
             radians(yaw), # Body yaw rate in radians
             thrust  # Thrust
         )
+
+    def _make_yaw_message(self, heading, yaw_speed=0, yaw_direction=1,relative=False):
+        return self.message_factory.command_long_encode(
+            0, 0,    # target system, target component
+            mavutil.mavlink.MAV_CMD_CONDITION_YAW, #command
+            0, #confirmation
+            heading,    # param 1, yaw in degrees
```
christopher-o-toole: what is this comment whitespace. make sure you're consistently using spaces and not tabs in your editor.

christopher-o-toole (changes requested):
Make sure that configurable values are defined as constants, particularly when they are used as default values in a function.

flight/tasks/yaw_task.py
```diff
@@ -0,0 +1,51 @@
...
+        self._yaw_direction = 1 #defaulted to 1 (clockwise)
+        self._relative = True #defaulted to always be relative
+
+    def perform(self):
+        """Do one iteration of logic for yawing the drone."""
+        if not self._has_started:
+            self.start_heading = self._drone.heading
+            self._drone.send_yaw(self._new_heading, self._yaw_speed, self._yaw_direction, self._relative)
+            self._has_started = True
+        #_is_not_finished stores whether the drone has finished the manuever within a certain degree of accuracy
```
ClayMav: There is no `_is_not_finished`

ClayMav: Just remove the comment

ClayMav (changes requested):
Codacy fixes required as well

ClayMav:
@jozgd Can we get this finished up?

ClayMav (approved):

christopher-o-toole (approved):

PR #33, opened 2019-02-27
Feature/task factory

Thomas-McKanna:
This pull request can be split into two sections

1. New argument parser for translating command line arguments into tasks
2. New TaskFactory class which can encode tasks into binary and decode them from binary into a fully configured Task object.

Note that the code in this request does not actually use the TaskFactory class, since it makes no sense to encode a message if it is not going to be sent over a network. However, I have tested the encoding and decoding functions to and they worked as intended in my testing.

The TaskFactory's methods are based off of an encoding which we have defined. How we are encoding is described by https://docs.google.com/spreadsheets/d/1evjB_2ChFrgppTaEZ03ktKQYPjTJT50ZoAbLgZen3No/edit#gid=0.


flight/drone/drone_controller.py
```diff
@@ -34,6 +37,8 @@ class DroneController(object):
 
     Attributes
     ----------
+    task_factory : flight.tasks.TaskFactory
```
ClayMav: I do not want the TaskFactory to be a part of the DroneController, it should only return a task object

flight/drone/drone_controller.py
```diff
@@ -86,8 +90,11 @@ def __init__(self, is_simulation=False):
             connection_string, wait_ready=True,
             heartbeat_timeout=c.CONNECT_TIMEOUT, status_printer=None,
             vehicle_class=Drone)
+        self._drone.is_simulation = is_simulation
         self._logger.info('Connected')
 
+        self.task_factory = TaskFactory(self._drone)
```
ClayMav: As above

flight/drone/drone_controller.py
```diff
@@ -16,7 +16,10 @@
 from flight import constants as c
 from drone import Drone
 import exceptions
-from flight.tasks import Hover, Takeoff, LinearMovement, Land, Exit, TakeoffSim
+
+import config
+from flight import constants as c
+from flight.tasks import Hover, Takeoff, LinearMovement, Land, Exit, TakeoffSim, TaskFactory
```
ClayMav: Most of these should not be required now, correct? If we are only passing tasks into the DroneController, then it shouldn't need to know how to make any of these except for Hover

ClayMav (changes requested):
First note, Tasks shouldn't need to have the drone object. If you can give me a good reason why it needs it, then we can talk about it.

Next, the way your constants are defined are very messy. There is a lot of duplicated information, and it is unclear why lots of it is needed. Your duplicated dictionaries with reversed keys and values could likely be solved with some kind of either list or ordered set.

In my stuff in network flight, I was using a class system to define each of the types of tasks, which in your case would give you the ability to store the ID of the task, the num_empty value, and order of parameters. The way you have it now is extremely messy, since you require a new function for every single task added. Look at the code I have in feature/networkflight in `flight/utils/messages.py` and `flight/utils/parser.py` to see how I did what I did, and try and do something similar, so that we can make this a lot more general.

We also discussed that this should not be making tasks, but a dictionary of keyword arguments

> {task: "linear_movement", direction:c.Directions.FORWARD, time: 10} etc.

The "Task Factory" will be constructed later on and will be used to take the above dictionary and turn it into a task, which will be how the drone will make tasks from here on out.

If you use the same method I used in feature/networkflight, you won't need to have a decode function for each either

Thomas-McKanna:
@ClayMav @christopher-o-toole Ready to be reviewed again.

flight/drone/drone.py
```diff
@@ -3,21 +3,26 @@
...
+import config
+from dronekit import connect, Vehicle, VehicleMode
 import logging
 from math import radians
 from pymavlink import mavutil
 import time
 
 from optical_flow_attribute import OpticalFlow
-from flight import constants as c
+from flight import constants as constants
```
ClayMav:
```suggestion
from flight import constants
```

flight/constants.py
```diff
@@ -3,6 +3,7 @@
 """
 
 from enum import Enum
+import numpy as np
```
ClayMav: Since this is only being used very few times, lets just use the full name here
```suggestion
import numpy
```

flight/constants.py
```diff
@@ -123,3 +124,9 @@ class MavBitmasks(Enum):
 # How many meters the drone can be away from target altitude before correcting
 # altitude
 ACCEPTABLE_ALTITUDE_DEVIATION = 0.1
+
+# Integer type used in encoding/decoding
+INT = np.int32
```
ClayMav:
```suggestion
INT = numpy.int32
```

flight/constants.py
```diff
@@ -123,3 +124,9 @@ class MavBitmasks(Enum):
...
+
+# Floating point type used in encoding/decoding
+FLOAT = np.float32
```
ClayMav:
```suggestion
FLOAT = numpy.float32
```

flight/tasks/decoder.py
```diff
@@ -0,0 +1,86 @@
+"""Class for decoding array of bytes into keyword dictionary."""
```
ClayMav, line 1: This should be in `flight/utils/decoder.py`

flight/tasks/decoder.py
```diff
@@ -0,0 +1,86 @@
...
+
+import numpy as np
+import sys
+import threading
+import traceback
+
+from encodings import Encodings
+from flight import constants as constants
```
ClayMav, line 9:
```suggestion
from flight import constants
```

flight/tasks/encodings.py
```diff
@@ -0,0 +1,103 @@
+"""A class used for encoding/decoding lookups."""
```
ClayMav, line 1: I will just say that I do not find this to be a very elegant solution. At this point there is no time to change it, but it is not very readable aside from the comments, and code should not only be readable by the inlayed comments. Also, the bidirectional dictionary is in my mind also not an elegant solution to the problem. I still believe the classed system to be the best approach, if only needing a little bit more work. Either way, this is the move for now, I suppose. Don't try to reinvent the wheel with new data structures if you don't have to.

flight/tasks/task_base.py
```diff
@@ -2,6 +2,9 @@
 
 # Standard Library
 import abc
+import uuid
```
ClayMav, line 5: I am not seeing where this is used

flight/utils/bidirectional_dictionary.py
```diff
@@ -0,0 +1,37 @@
+"""A dictionary and can be searched by key or by value."""
```
ClayMav, line 1: Really don't like this approach from a usability standpoint, no need to overcomplicate it.

testing_cli.py
```diff
@@ -1,202 +1,235 @@
...
+import config
 from flight.drone.drone_controller import DroneController
-from flight import constants as c
+from flight import constants
+from flight.tasks import TaskFactory
+from flight.tasks.encodings import Encodings
 
 PROMPT_FOR_COMMAND = '> '
 
+CLI_STR_TO_TASK = {
```
ClayMav: All of this should definitely be defined in a grammar somewhere else, a system similar to my attempt in networkflight could solve this since each class was named. The fact that we have multiple definitions mapping x to y and y to x is wasteful and confusing.

Again, I think you were trying to reinvent the wheel

testing_cli.py
```diff
@@ -1,202 +1,235 @@
...
+    # Hover sub-parser
+    parser_hover = subparsers.add_parser('hover', help='hover help')
+    parser_hover.add_argument('--duration',
+        dest='duration',
+        action=store_float(),
+        required=True,
+        help='altitude to maintain during hover')
+    parser_hover.add_argument('--altitude',
+        dest='altitude',
+        action=store_float(),
```
ClayMav, line 119: Pretty sure you can just set a type for the argument

testing_cli.py
```diff
@@ -1,202 +1,235 @@
...
+        dest='altitude',
+        default=config.DEFAULT_ALTITUDE,
+        action=store_float(),
+        help='altitude to maintain during move')
+
+    # Hover sub-parser
+    parser_hover = subparsers.add_parser('hover', help='hover help')
+    parser_hover.add_argument('--duration',
+        dest='duration',
+        action=store_float(),
```
ClayMav, line 114: Pretty sure you can just set a type for the argument

testing_cli.py
```diff
@@ -1,202 +1,235 @@
...
+        help='altitude to maintain during move')
+    parser_linear_move.add_argument('--direction',
+        dest='direction',
+        action=store_direction(),
+        required=True,
+        help='direction to move in')
+    parser_linear_move.add_argument('--altitude',
+        dest='altitude',
+        default=config.DEFAULT_ALTITUDE,
+        action=store_float(),
```
ClayMav, line 106: Pretty sure you can just set a type for the argument

testing_cli.py
```diff
@@ -1,202 +1,235 @@
...
+        dest='altitude',
+        action=store_float(),
+        default=config.DEFAULT_ALTITUDE,
+        help='altitude to take off to')
+
+    # Linear movement sub-parser
+    parser_linear_move = subparsers.add_parser('move', help='move help')
+    parser_linear_move.add_argument('--duration',
+        dest='duration',
+        action=store_float(),
```
ClayMav, line 95: Pretty sure you can just set a type for the argument

testing_cli.py
```diff
@@ -1,202 +1,235 @@
...
+    parser_exit = subparsers.add_parser('exit', help='exit help')
+
+    # Land sub-parser (has no arguments)
+    parser_land = subparsers.add_parser('land', help='land help')
+
+    # Takeoff sub-parser
+    parser_takeoff = subparsers.add_parser('takeoff', help='takeoff help')
+    parser_takeoff.add_argument('--altitude',
+        dest='altitude',
+        action=store_float(),
```
ClayMav, line 86: Pretty sure you can just set a type for the argument

testing_cli.py
```diff
@@ -1,202 +1,235 @@
...
-                print('Closing session.')
-                return
-
-            print('{}{}: {}'.format(PROMPT_FOR_COMMAND, type(e).__name__, e))
+            exc_type, exc_value, exc_traceback = sys.exc_info()
+            traceback.print_exception(exc_type, exc_value, exc_traceback,
+                                        limit=2, file=sys.stdout)
 
 
 def get_priority(priority):
```
ClayMav, line 197: priority and direction use the same code

ClayMav (changes requested):
Lot of code so hopefully I caught everything. Codacy also has complaints.

ClayMav:
Closing to make way for new pull requests with better defined structure.

PR #34, opened 2019-02-28
Removed enums, using base classes instead

ClayMav:


Thomas-McKanna (approved):
Looks good

christopher-o-toole (commented):
The benefit of using enums instead of classes with static member variables is that it allows for type checking and validation, so I'm not sure that getting rid of enums is worth not having to add the .value accessor. Also for debugging purposes, it's a lot more informative to print enum values than arbitrary constant values.

ClayMav:
This PR is fixing an error, as different installations of the code were sometimes failing or not failing. We weren't using the benefits anyways, so we thought to get rid of it. I stand with that decision to solve stability between installs

christopher-o-toole (changes requested):
Not using the benefits of something is not a justification for removing it. Add any relevant packages to the  `requirements.txt` to fix any installation issues, and note any other necessary modifications in the readme or a dedicated install guide.

PR #35, opened 2019-02-28
Changed the long hover to hover at the drone's current altitude

jozgd:
Now the drone will long hover at its current altitude when it runs out of tasks

### DIFF
--- a/flight/drone/drone_controller.py
+++ b/flight/drone/drone_controller.py
@@ -249,7 +249,7 @@ class DroneController(object):
         # If there are no more tasks, begin to hover.
         if self._drone.armed and self._current_task is None:
             self._logger.info('No more tasks - beginning long hover')
-            self.add_hover_task(config.DEFAULT_ALTITUDE,
+            self.add_hover_task(self._drone.rangefinder.distance,
                                 c.DEFAULT_HOVER_DURATION)
 
         return True
--- a/flight/drone/drone_controller.py
+++ b/flight/drone/drone_controller.py
@@ -152,7 +152,7 @@ class DroneController(object):
         priority : Priorities.{LOW, MEDIUM, HIGH}, optional
             The importance of this task.
         """
-        if altitude = None:
+        if altitude == None:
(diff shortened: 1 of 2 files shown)


ClayMav (changes requested):
I thought we had discussed being able to create hover tasks with no altitude and they would default to the current altitude. Then this line would just not include a height.

Thomas-McKanna:
I know I told you this would work, but I wasn't thinking of the larger picture. Clay is right in that if the hover task is given no altitude it should default to its current altitude. There could be instances other than when we run out of tasks where we want to make a hover task that remains at current altitude, and we don't want to have to pass in the current altitude as a parameter each time.

I would make altitude an optional argument that defaults to None and then grab the current altitude at the first iteration of perform.

christopher-o-toole (dismissed):
Looks fine after the latest commit.

flight/drone/drone_controller.py
```diff
@@ -140,7 +140,7 @@ def run(self):
             sleep(c.DELAY_INTERVAL)  # Sleep in case was doing write operation
             self._splitter.exit()
 
-    def add_hover_task(self, altitude, duration, priority=c.Priorities.LOW):
+    def add_hover_task(self, duration=c.DEFAULT_HOVER_DURATION, altitude=None, priority=c.Priorities.LOW):
```
ClayMav, line 144: Just default altitude to be rangefinder distance, since that's the only checking you are doing right now

ClayMav (changes requested):

ClayMav:
@jozgd Can we get this finished up as well?

ClayMav:
Fixes #55

Thomas-McKanna (approved):
I have gone through the code with John and we tested it in the simulator. It seems to be working (we did notice some anomalies in the simulator, such as occasionally drifting slowly downward). I think that this weird behavior was just the simulator acting up, since the changed lines of code in this pull request do not affect changing the altitude, only setting a target altitude.

ClayMav (approved):

PR #36, opened 2019-02-28
Pitch and roll exceptions

BenjaminWantland:
I added two new exceptions for pitch and roll. There is a constant to change the range of the exception.

### DIFF
--- a/flight/constants.py
+++ b/flight/constants.py
@@ -123,3 +123,6 @@ OPTICAL_FLOW_MESSAGE = 'OPTICAL_FLOW'
 # How many meters the drone can be away from target altitude before correcting
 # altitude
 ACCEPTABLE_ALTITUDE_DEVIATION = 0.1
+
+# Maximum roll and pitch (in degrees) before being considered unsafe
+MAXIMUM_PITCH_ROLL = 45
--- a/flight/drone/drone_controller.py
+++ b/flight/drone/drone_controller.py
@@ -12,6 +12,7 @@ from threading import Event
 from time import sleep
 import traceback
 
+from math import radians
 import config
 from flight import constants as c
 from drone import Drone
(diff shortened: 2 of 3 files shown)


ClayMav:
Could not replicate this in the sim, and I hope we never do, @Thomas-McKanna can you think of a good way to validate this works?

Thomas-McKanna (changes requested):
Everything is in place, but this is one big problem: pitch, roll, and yaw are given in radians, not degrees. 

The way I figured this out was by testing the code in my simulator and printing out the pitch and roll values. I noticed they were pretty close to 0 almost all the time and moving only bumped it up to roughly 0.1.

Then I looked at the Mavlink message structure for attitude at https://mavlink.io/en/messages/common.html#ATTITUDE, and there you can see that the units are in radians. All of the data we can get through Dronekit comes from Mavlink messages, so they are typically a good place to check for stuff like this.

Change the constants to radians and I would set the threshold at 0.2 for now (that's double the highest reading I saw in the sim) If it is too low we can always bump it up.

BenjaminWantland:
@Thomas-McKanna 
The constant is set in degrees but it is converted to radians before it checks. I just thought the const would be easier to change if in degrees.

Thomas-McKanna:
A ha! I did not see that. Works for me.

Thomas-McKanna (dismissed):

Thomas-McKanna:
@ClayMav If you want to test it, you could set the threshold to something very low like 2 degrees and see if it triggers then.

ClayMav:
Alright, I am convinced this is alright thanks to Thomas, Can you fix the codacy problems for me, then we can get this finished.

ClayMav (changes requested):
As above

ClayMav (approved):

Thomas-McKanna (approved):

PR #68, opened 2019-05-23
QR Code generator

coledie:
A qr code generator class able to create full codes and just sections of qr codes.


vision/qr_gen/QrCode.py
```diff
@@ -0,0 +1,177 @@
...
+    @encoded_value.setter
+    def encoded_value(self, value):
+        """
+        Sets the value represented by the QR code and updates all fields to represent this change.
+        """
+
+        self._encoded_value = value
+        qr = qrcode.QRCode(
+            version=None,
+            error_correction=qrcode.constants.ERROR_CORRECT_L,
```
ClayMav: Should be H

vision/qr_gen/QrCode.py
```diff
@@ -0,0 +1,177 @@
...
+
+    @encoded_value.setter
+    def encoded_value(self, value):
+        """
+        Sets the value represented by the QR code and updates all fields to represent this change.
+        """
+
+        self._encoded_value = value
+        qr = qrcode.QRCode(
+            version=None,
```
ClayMav: Should be one

ClayMav (changes requested):
I want to know for a fact that this generates equivalent qr codes to what it should, and fix the codacy errors

vision/qr_gen/requirements.txt
```diff
@@ -0,0 +1,4 @@
+qrcode
+opencv-python
+numpy
+qrcode
```
ClayMav: Duplicate entry
```suggestion
```

ClayMav (changes requested):

PR #69, opened 2019-05-25
Update to new format

ClayMav:


ClayMav:
Now outscaled by feature/swarming_test

PR #72, opened 2019-06-05
Compile accumulator to dll. Python to interact with dll.

coledie:


ClayMav (dismissed):
I would appreciate if you documented where you are getting some of your equations and what not. Otherwise, this will be very hard for future people to pick through.

ClayMav (approved):

